PR #1, opened 2023-07-12
Updating data API example

marcus-girolneto:


examples/DataApi-Water-And-Energy-Usage/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -12,69 +12,218 @@ import org.http4s.HttpRoutes
...
+    else
+      EnergyNotPositive.asInstanceOf[DataApplicationValidationError].invalidNec
+
+  def validateWaterUsageUpdate(water: WaterUsage): DataApplicationValidationErrorOr[Unit] =
+    if (water.usage > 0L && water.timestamp > 0L)
+      ().validNec
+    else
+      WaterNotPositive.asInstanceOf[DataApplicationValidationError].invalidNec
+
+  def validateEnergyTimestampToNoDoubling(maybeState: Option[AggregatedUsage], energy: EnergyUsage): DataApplicationValidationErrorOr[Unit] = {
```
AlexBrandes: I think the intention here with this method name is to indicate that the validation prevents block doubling/double spends but I don't think anyone would understand that without more context. I would suggest changing this name to something more readable like `validateEnergyTimestamp` and leaving a comment that explains the implication of requiring a greater timestamp than the previous update.

examples/DataApi-Water-And-Energy-Usage/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -12,69 +12,218 @@ import org.http4s.HttpRoutes
...
+  case object EnergyAlreadyUpdated extends DataApplicationValidationError {
+    val message = "Energy already updated, skipping"
+  }
+
+  case object WaterNotPositive extends DataApplicationValidationError {
+    val message = "water usage must be positive"
+  }
+
+  case object WaterAlreadyUpdated extends DataApplicationValidationError {
+    val message = "Energy already updated, skipping"
```
AlexBrandes:
```suggestion
    val message = "Water already updated, skipping"
```

examples/DataApi-Water-And-Energy-Usage/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -12,69 +12,218 @@ import org.http4s.HttpRoutes
...
 
 import java.nio.charset.StandardCharsets
 
+object Errors {
+  case object EnergyNotPositive extends DataApplicationValidationError {
+    val message = "energy usage must be positive"
+  }
+
+  case object EnergyAlreadyUpdated extends DataApplicationValidationError {
+    val message = "Energy already updated, skipping"
```
AlexBrandes: What's actually happening in this case is that the Energy update is older than the most recent update so it's rejected. How about `EnergyUpdateOutdated` with message "Energy update older than latest update timestamp, rejecting" 

+ similar pattern for Water

examples/DataApi-Water-And-Energy-Usage/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -12,69 +12,218 @@ import org.http4s.HttpRoutes
...
+    for {
+      validatedUpdates <- validateUpdates
+      validatedAddress <- validateAddress
+      validatedTimestamp <- validateTimestamp
+    } yield validatedUpdates.productR(validatedAddress).productR(validatedTimestamp)
+
   }
 
-  def combine(oldState: State, updates: NonEmptyList[Signed[Usage]]): IO[State] =
+  def validateUpdate(update: Update): IO[DataApplicationValidationErrorOr[Unit]] = {
```
AlexBrandes: I think we should have descriptive comments for each of these major methods (validateData, validateUpdate, combine, serialize*) describing what their role is and what we're doing in the code. This is an example app so it would be helpful to have much more verbose comments than usual.

examples/DataApi-Water-And-Energy-Usage/template/modules/l0/src/main/scala/com/my/currency/l0/Main.scala
```diff
@@ -23,29 +23,32 @@ object Main
     ClusterId(UUID.fromString("517c3a05-9219-471b-a54c-21b7d72f4ae5")),
     version = BuildInfo.version
   ) {
-  def dataApplication:  Option[BaseDataApplicationL0Service[IO]] = Option(BaseDataApplicationL0Service(new DataApplicationL0Service[IO, Usage, State] {
-    override def genesis: State = State(0,0)
+  def dataApplication: Option[BaseDataApplicationL0Service[IO]] =
+    Option(BaseDataApplicationL0Service(new DataApplicationL0Service[IO, Update, State] {
 
-    override def validateData(oldState: State, updates: NonEmptyList[Signed[Usage]])(implicit context: L0NodeContext[IO]): IO[DataApplicationValidationErrorOr[Unit]] = Data.validateData(oldState, updates)
+    override def genesis: State = State(Map.empty)
```
AlexBrandes, line 39: Can we call this out with a comment? `State` is a pretty important thing to understand if trying to extend this example.

examples/DataApi-Water-And-Energy-Usage/README.md
```diff
@@ -1,57 +1,83 @@
-
 # Metagraph - Data API - Water and Energy Usage
 
-  
-
 This is one example of the Data API, where we can send transactions to update the state of Water and Energy usage
```
AlexBrandes:
```suggestion
This example uses the Data API to demonstrate a basic IoT use case. In the example a client, representing a sensor, sends signed data updates to a metagraph which validates the data before merging the updates into snapshot state. 
```

examples/DataApi-Water-And-Energy-Usage/README.md
```diff
@@ -1,57 +1,83 @@
...
 
+`modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala`
 
 ### shared_data
-->   `validateData` and `validateUpdate`
-* These functions will check the provided data to verify its validity. If the amount that we want to update  of water and energy usage is valid (at this example > 0), we will proceed to the state update, otherwise, we will return an error
+
+-> `validateData`
+
+- This function will check the provided data to verify its validity. This validation happens on the L0 layer. In this case, we consider the `oldState` and the `proofs`. For this example, we check if the provided address is the same that the address that signed the message. We also check the `oldState` to avoid doubling when updating
```
AlexBrandes:
```suggestion
- This function will check the provided data to verify its validity. This validation happens on the L0 layer. In this case, we consider the `oldState` and the `proofs`. For this example, we check that the provided address is the same as the address that signed the message. We also check the most recent update timestamp in order to prevent old data or duplicated data from being accepted.
```

examples/DataApi-Water-And-Energy-Usage/README.md
```diff
@@ -1,57 +1,83 @@
...
+
+-> `validateUpdate`
+
+- Pretty similar to the `validateData` method, but this one will happen on the L1 layer and will do the initial validations that don't need to consider the `proofs` and the `oldState`. In this case, we're validating if the provided usage is greater than 0.
```
AlexBrandes:
```suggestion
- This method performs initial validation on the update on the L1 layer. It does not have access to state information. 
```

Also, I'd suggest putting this method above `validateData` because it happens first in the sequence.

examples/DataApi-Water-And-Energy-Usage/README.md
```diff
@@ -1,57 +1,83 @@
...
 
 -> `combine`
-* This function will combine the old state with the new update provided. In this case, if we provide `EnergyUsage` of 10, we will sum 0 + 10
 
--> `serializeState` and `serializeUpdate`
-* These are the functions that will serialize the JSON. In this case, one example of request that is serialized for these functions:
- ```
- {"value":{"EnergyUsage":{"value":10}},"proofs":[{"id":"e75a6011eaa38d7b0a1cb41810c655cdc89c6c5ffd207cbab9d18fd49cbf2729e262b5387a4687a23a163d14bc0dff8ef6539e2a73932e77d2de6b1895facd99","signature":"3044022060107a64dabbc9b0e2779a9fada99646798b3ebd21ecda22c7de2740f13addc30220564d8303f1a581f9a7c021252f624911b7c3bec63620a0d000b7f651753e020d"}]}
- ```
+- This function will combine the old state with the new update provided. In this case, if we provide `EnergyUsage` of 10, we will sum 0 + 10
```
AlexBrandes:
```suggestion
- This method accepts validated data and the previous state which is combined to return the new state. 
```

examples/DataApi-Water-And-Energy-Usage/README.md
```diff
@@ -1,57 +1,83 @@
...
+
+-> `serializeState`
+
+- These are the functions that will serialize the state JSON. In this case, one example of request that is serialized for these functions:
```
AlexBrandes: This method serializes state not an update so the example doesn't make sense.

examples/DataApi-Water-And-Energy-Usage/README.md
```diff
@@ -1,57 +1,83 @@
...
+
+```
+
+{"value":{"EnergyUsage":{"value":10}},"proofs":[{"id":"e75a6011eaa38d7b0a1cb41810c655cdc89c6c5ffd207cbab9d18fd49cbf2729e262b5387a4687a23a163d14bc0dff8ef6539e2a73932e77d2de6b1895facd99","signature":"3044022060107a64dabbc9b0e2779a9fada99646798b3ebd21ecda22c7de2740f13addc30220564d8303f1a581f9a7c021252f624911b7c3bec63620a0d000b7f651753e020d"}]}
+
+```
+
+-> `serializeUpdate`
+
+- These are the functions that will serialize the update JSON. In this case, one example of request that is serialized for these functions:
```
AlexBrandes:
```suggestion
- This is the method that serializes the update JSON into an internal format. In this case, one example of request that is serialized for these functions:
```

AlexBrandes (changes requested):
Overall, I think this is a great example. Adding some comments and cleaning up the existing description would help with developer understanding.

AlexBrandes (approved):

PR #2, opened 2023-07-17
Updated README and Data API example comments

AlexBrandes:


marcus-girolneto (approved):
Looks perfect, great updates on the comments and README, way more legible now thanks for that ✌🏼

PR #3, opened 2023-07-19
Adding Poll Example - Initial commit

marcus-girolneto:


examples/DataApi-Voting-poll/scripts/send_data_transaction_create_poll.js
```diff
@@ -0,0 +1,101 @@
+const { dag4 } = require( '@stardust-collective/dag4' );
```
AlexBrandes: Can we put all of the sending functions into a single file? There's a lot of repeated code otherwise.

examples/DataApi-Voting-poll/scripts/send_data_transaction_vote_poll.js
```diff
@@ -0,0 +1,98 @@
...
+
+const sha256 = ( hash ) => {
+    return jsSha256.sha256( hash );
+};
+
+const sha512 = ( hash ) => {
+    return jsSha512.sha512( hash );
+};
+
+const sign = async ( privateKey, msg ) => {
```
AlexBrandes: Does `dag4.keystore.sign` work here? Either that or we should migrate this to a library somewhere.

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/Main.scala
```diff
@@ -0,0 +1,65 @@
...
+  extends CurrencyL0App(
+    "currency-l0",
+    "currency L0 node",
+    ClusterId(UUID.fromString("517c3a05-9219-471b-a54c-21b7d72f4ae5")),
+    version = BuildInfo.version
+  ) {
+  def dataApplication: Option[BaseDataApplicationL0Service[IO]] =
+    Option(BaseDataApplicationL0Service(new DataApplicationL0Service[IO, PollUpdate, State] {
+
+    /**
```
AlexBrandes: Can remove this comment

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,351 @@
...
+        }
+      case None => PollDoesNotExists.asInstanceOf[DataApplicationValidationError].invalidNec
+    }
+  }
+}
+
+object Routes {
+  def routeValidateUpdate(update: PollUpdate, state: State, lastSnapshotOrdinal: SnapshotOrdinal): IO[DataApplicationValidationErrorOr[Unit]] = {
+    update match {
+      case newPoll: NewPoll =>
```
AlexBrandes: This sort of defeats the purpose of "routing" the validations since we still have all the code for each validated type in a single method. The idea would be to organize this code a bit more by splitting the validation for each type into its own function.

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,351 @@
...
+}
+
+object Data {
+  @derive(decoder, encoder)
+  sealed trait PollUpdate extends DataUpdate {
+    val timestamp: Long
+  }
+
+  @derive(decoder, encoder)
+  case class NewPoll(name: String, owner: Address, pollOptions: List[String], startSnapshotOrdinal: Long, endSnapshotOrdinal: Long, timestamp: Long) extends PollUpdate
```
AlexBrandes: I would avoid using "New" in a method or variable name anywhere since it's a keyword in Scala and most languages. Maybe `createPollUpdate` is better?

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,351 @@
...
+
+  @derive(decoder, encoder)
+  case class PollVote(id: String, address: Address, option: String, timestamp: Long) extends PollUpdate
```
AlexBrandes: Maybe `voteInPollUpdate` is better?

Also, is `id` here referring to the ID of the poll? or of the vote? If it's of the poll then I think it should be `pollId` instead of `id`.

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,351 @@
...
+      json.as[PollUpdate]
+    }
+  }
+}
+
+object Data {
+  @derive(decoder, encoder)
+  sealed trait PollUpdate extends DataUpdate {
+    val timestamp: Long
+  }
```
AlexBrandes: I guess there's a requirement that NewPoll and PollVote both extend a common trait. Can this trait be empty? I don't think `timestamp` makes sense in either update. Maybe `id` or `hash` instead?

examples/DataApi-Voting-poll/scripts/send_data_transaction_create_poll.js
```diff
@@ -0,0 +1,101 @@
...
+
+    account1.connect( {
+        networkVersion: '2.0',
+        l0Url: globalL0Url,
+        testnet: true
+    } );
+
+    const message = {
+        NewPoll: {
+            name: ':poll_name',
```
AlexBrandes: Is `name` supposed to be ID? Can we set the ID based on the hash of the poll that's created?

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,351 @@
...
+          ClosedPool.asInstanceOf[DataApplicationValidationError].invalidNec
+        } else {
+          ().validNec
+        }
+      case None => PollDoesNotExists.asInstanceOf[DataApplicationValidationError].invalidNec
+    }
+  }
+}
+
+object Routes {
```
AlexBrandes: We probably shouldn't name this object `Routes` to avoid confusion with HTTP routes.

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,351 @@
...
+          acc.focus(_.polls).modify(_.updated(voteId.toString, newState))
+        case pollVote: PollVote =>
+          val currentState = acc.polls(pollVote.id)
+          val currentOptionNumber = currentState.pollOptions(pollVote.option)
+
+          val newState = currentState
+            .focus(_.pollOptions)
+            .modify(_.updated(pollVote.option, currentOptionNumber + 1))
+            .focus(_.votes)
+            .modify(_.updated(pollVote.address, 1))
```
AlexBrandes: This needs to record what option the address voted for.

AlexBrandes (changes requested):

examples/DataApi-Voting-poll/scripts/send_data_transaction_create_poll.js
marcus-girolneto: Done

examples/DataApi-Voting-poll/scripts/send_data_transaction_vote_poll.js
marcus-girolneto: Done

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
marcus-girolneto: Updated

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
marcus-girolneto: It can be empty

examples/DataApi-Voting-poll/scripts/send_data_transaction_create_poll.js
marcus-girolneto: No, the name it's just the poll name. I've added the ID as well now

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
marcus-girolneto: Changed to pollId

examples/DataApi-Voting-poll/scripts/send_data_transaction_create_poll.js
marcus-girolneto: The id of the poll is exactly the hash returned from the `/data` endpoint

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,360 @@
...
+
+  def combineVoteInPoll(voteInPoll: VoteInPollUpdate, acc: State): State = {
+    val currentState = acc.polls(voteInPoll.pollId)
+    val currentOptionNumber = currentState.pollOptions(voteInPoll.option)
+
+    val newState = currentState
+      .focus(_.pollOptions)
+      .modify(_.updated(voteInPoll.option, currentOptionNumber + 1))
+      .focus(_.votes)
+      .modify(_.updated(voteInPoll.address, Map(voteInPoll.option -> 1)))
```
AlexBrandes: It looks like this sets the `voteInPoll.option` value to 1. Shouldn't it add 1 to the current value?

marcus-girolneto: Oh sorry, the name is confusing, This only sets that the user voted in this option, currently, the user can only vote 1 in the option. The option above pollOptions that we're increasing in 1 is the part you mentioned

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,360 @@
...
+import monocle.syntax.all._
+import org.tessellation.schema.SnapshotOrdinal
+import org.tessellation.schema.address.Address
+import org.tessellation.security.SecurityProvider
+import org.tessellation.security.hash.Hash
+import org.tessellation.security.signature.signature.SignatureProof
+
+import java.nio.charset.StandardCharsets
+
+object Errors {
```
ryle-goehausen: This object can be put in its own Errors file

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,360 @@
...
+    val message = "Could not get latest state!"
+  }
+
+  case object ClosedPool extends DataApplicationValidationError {
+    val message = "Pool is closed"
+  }
+}
+
+object TypeValidators {
+  def validateIfPollAlreadyExists(maybeState: Option[PollState]): DataApplicationValidationErrorOr[Unit] = {
```
ryle-goehausen: This validation function and the one below it are a little confusing in that they're only ever dealing with the same option type. It might make more sense to implement this logic:

`validateIfPollAlreadyExists(state.polls.get(voteId.toString))`

from the invocation call of this, directly in the validate function by passing State and voteId as parameters. Since this isn't used anywhere else (i.e. the result of the get is not used anywhere else,) it makes sense to implement it only in the validator.

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,360 @@
...
+
+    for {
+      validatedCreatePollSnapshot <- validateCreatePollSnapshot
+      validatedPoll <- validatePoll
+    } yield validatedCreatePollSnapshot.productR(validatedPoll)
+
+  }
+
+  def voteInPollValidations(update: VoteInPollUpdate, state: State, lastSnapshotOrdinal: Option[SnapshotOrdinal]): IO[DataApplicationValidationErrorOr[Unit]] = {
+    val validateSnapshotInterval = IO {
```
ryle-goehausen: For a lot of other places in the code, we're constructing an IO type here to be consistent with outer functions that require an F type. Since it appears most of the uses here don't actually require an IO in the sense that they're not dealing with a side effect like reading a file (and are instead doing something instantaneous like checking a map or applying a comparison,) it's not really appropriate to use an IO for that. Even if we wanted to compose them, it would be better to use an arbitrary F type, but I wouldn't recommend that either.

Instead, what you should do here is just process the simple logic as is here (without any IO types,) and only at the immediate point that you require it to be used for compatibility with the outer functions, then construct an IO there.

You can also use IO.pure if it's an immediate calculated value (instead of a longer running function which is what the thunk is for.) I.e. if the value has ALREADY been materialized (for instance invoking X > 5 and then assigning to a value,) you can use pure, if the thunk you're passing to IO.apply here (the syntax for (IO{...}) is an actual calculation, then its appropriate to use IO.apply -- but not for already calculated values

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,360 @@
...
+
+    val validateOption = IO {
+      validateIfOptionExists(state.polls.get(update.pollId), update.option)
+    }
+
+    val validateRepeatedVote = IO {
+      validateIfUserAlreadyVoted(state.polls.get(update.pollId), update.address)
+    }
+
+    for {
```
ryle-goehausen: As a note, it appears most of these underlying functions don't require IO, hence we're constructing an IO here and immediately doing a for comprehension flatmap syntax to remove them, making this somewhat redundant versus just using productR directly on the final immediate types.

The only reason to bind something into an IO should be for interaction with another IO already present, which most of these functions do not appear to have.

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,360 @@
...
+  }
+
+  def customUpdateDeserialization(bytes: Array[Byte]): Either[Throwable, PollUpdate] = {
+    parser.parse(new String(bytes, StandardCharsets.UTF_8)).flatMap { json =>
+      json.as[PollUpdate]
+    }
+  }
+}
+
+object Data {
```
ryle-goehausen: I would preserve this portion as the main object in this Main method here, since it most closely encapsulates the logic repeated elsewhere, and move everything else to different utility / error / validation files.

ryle-goehausen (changes requested):

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
marcus-girolneto: Done

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
marcus-girolneto: Done

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
marcus-girolneto: Thanks for the tip man, updated

ryle-goehausen (approved):
LGTM, Alex should double check to confirm. 

Only remaining points of concern are about extends DataState, I assume this will be used in the future for other extensions? It seems right now we have just the state extending it and the poll extending it but no real usage of it, maybe I missed that?

There's some other IO cleanup that can be done here, but mostly seems fixed. Not as much of a priority.

PR #6, opened 2023-08-18
Add sample NFT UI app

juandavidkincaid:
### Changes
+ Added new basic CRA app for NFT example.
+ Added utility scripts to interact with the metagraph.

### Notes
+ Due to CORS restrictions, the app will redirect traffic to the DataAPI on the same port it runs.


marcus-girolneto (approved):
You only added a new directory with the sample UI, so IMO it's great since the UI had worked well at the example

examples/DataApi-NFT/sample-ui/README.md
```diff
@@ -0,0 +1,46 @@
+# Getting Started with Create React App
```
AlexBrandes: Can we update this README to have some detail about this FE app and how to run it? It would be good to have some instructions on how to run the scripts, etc.

marcus-girolneto: I agree with Alex, just add a new section on README with basic running instructions

PR #7, opened 2023-08-22
Updating energy and water usage example

marcus-girolneto:
### Changes
+ Refactored example to use the new files structure
+ Added chaining to list the transactions
Captura de Tela 2023-08-22 às 15 43 20


examples/DataApi-Water-And-Energy-Usage/template/modules/data_l1/src/main/scala/com/my/currency/data_l1/CustomRoutes.scala
```diff
@@ -0,0 +1,94 @@
...
+  private def getState(context: L1NodeContext[IO]) = {
+    OptionT(context.getLastCurrencySnapshot)
+      .flatMap(_.data.toOptionT)
+      .flatMapF(deserializeState(_).map(_.toOption))
+      .value
+  }
+
+  private def getMetagraphStateFromSnapshotOrdinal(ordinal: Long): Array[Byte] = {
+    val headers: Iterable[(String, String)] = Iterable(("Accept", "application/json"))
+    val response = requests.get(s"http://host.docker.internal:9400/snapshots/$ordinal", headers = headers)
```
AlexBrandes, line 26: Is there an internal method we can use to fetch this instead of making an HTTP call?

marcus-girolneto: I don't know, @ryle-goehausen do you know if we have something like this?

ryle-goehausen: Just to clarify again from discussion only the L0 could in theory access this data, so on the L1 this is probably best way to accomplish this, route could potentially be moved

ryle-goehausen (approved):
I think this is correct, had a concern about the combiners I was looking into regarding what happens on failure but I don't think it applies here.

PR #8, opened 2023-11-16
Updating vote poll example

marcus-girolneto:
* Updating the vote poll example to work with changes made by inclusion of calculatedState


examples/DataApi-Voting-poll/template/modules/data_l1/src/main/scala/com/my/currency/data_l1/Main.scala
```diff
@@ -26,35 +30,45 @@ object Main
     ClusterId(UUID.fromString("517c3a05-9219-471b-a54c-21b7d72f4ae5")),
     version = BuildInfo.version
   ) {
-  override def dataApplication: Option[BaseDataApplicationL1Service[IO]] = Option(BaseDataApplicationL1Service(new DataApplicationL1Service[IO, PollUpdate, State] {
-
-    override def serializeState(state: State): IO[Array[Byte]] = MainData.serializeState(state)
+  override def dataApplication: Option[BaseDataApplicationL1Service[IO]] = Option(BaseDataApplicationL1Service(new DataApplicationL1Service[IO, PollUpdate, VoteStateOnChain, VoteCalculatedState] {
+    override def validateUpdate(update: PollUpdate)(implicit context: L1NodeContext[IO]): IO[DataApplicationValidationErrorOr[Unit]] = shared_data.Main.validateUpdate(update)
```
sdavidp: Calling a method in `Main` ? This looks very odd...

examples/DataApi-Voting-poll/template/modules/data_l1/src/main/scala/com/my/currency/data_l1/Main.scala
```diff
@@ -26,35 +30,45 @@ object Main
...
 
-    override def deserializeState(bytes: Array[Byte]): IO[Either[Throwable, State]] = MainData.deserializeState(bytes)
+    override def validateData(state: DataState[VoteStateOnChain, VoteCalculatedState], updates: NonEmptyList[Signed[PollUpdate]])(implicit context: L1NodeContext[IO]): IO[DataApplicationValidationErrorOr[Unit]] = IO {
+      ().validNec
+    }
```
sdavidp:
```suggestion
   override def validateData(state: DataState[VoteStateOnChain, VoteCalculatedState], updates: NonEmptyList[Signed[PollUpdate]])(implicit context: L1NodeContext[IO]): IO[DataApplicationValidationErrorOr[Unit]] = ().validNec.pure[IO]
```
Also, it looks odd, esp. if this is going to serve as an example, that we are returning a validNec when the function is typed for `DataApplicationValidationErrorOr[Unit]` because it means the reader has to have knowledge of what that return type is defined as, similar to knowing an implementation of an interface in order to use it correctly.

examples/DataApi-Voting-poll/template/modules/data_l1/src/main/scala/com/my/currency/data_l1/Main.scala
```diff
@@ -26,35 +30,45 @@ object Main
     ClusterId(UUID.fromString("517c3a05-9219-471b-a54c-21b7d72f4ae5")),
     version = BuildInfo.version
   ) {
-  override def dataApplication: Option[BaseDataApplicationL1Service[IO]] = Option(BaseDataApplicationL1Service(new DataApplicationL1Service[IO, PollUpdate, State] {
-
-    override def serializeState(state: State): IO[Array[Byte]] = MainData.serializeState(state)
+  override def dataApplication: Option[BaseDataApplicationL1Service[IO]] = Option(BaseDataApplicationL1Service(new DataApplicationL1Service[IO, PollUpdate, VoteStateOnChain, VoteCalculatedState] {
```
sdavidp, line 33: FYI, the return type for this hook may be changing.

examples/DataApi-Voting-poll/template/modules/data_l1/src/main/scala/com/my/currency/data_l1/Main.scala
```diff
@@ -26,35 +30,45 @@ object Main
...
 
-    override def deserializeState(bytes: Array[Byte]): IO[Either[Throwable, State]] = MainData.deserializeState(bytes)
+    override def validateData(state: DataState[VoteStateOnChain, VoteCalculatedState], updates: NonEmptyList[Signed[PollUpdate]])(implicit context: L1NodeContext[IO]): IO[DataApplicationValidationErrorOr[Unit]] = IO {
+      ().validNec
+    }
 
-    override def serializeUpdate(update: PollUpdate): IO[Array[Byte]] = MainData.serializeUpdate(update)
+    override def combine(state: DataState[VoteStateOnChain, VoteCalculatedState], updates: List[Signed[PollUpdate]])(implicit context: L1NodeContext[IO]): IO[DataState[VoteStateOnChain, VoteCalculatedState]] = IO {
+      state
+    }
```
sdavidp:
```suggestion
    override def combine(state: DataState[VoteStateOnChain, VoteCalculatedState], updates: List[Signed[PollUpdate]])(implicit context: L1NodeContext[IO]): IO[DataState[VoteStateOnChain, VoteCalculatedState]] = state.pure[IO]
```

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/Main.scala
```diff
@@ -26,31 +36,50 @@ object Main
...
   def dataApplication: Option[BaseDataApplicationL0Service[IO]] =
-    Option(BaseDataApplicationL0Service(new DataApplicationL0Service[IO, PollUpdate, State] {
-      override def genesis: State = State(Map.empty)
+    Option(BaseDataApplicationL0Service(new DataApplicationL0Service[IO, PollUpdate, VoteStateOnChain, VoteCalculatedState] {
+      override def genesis: DataState[VoteStateOnChain, VoteCalculatedState] = DataState(VoteStateOnChain(List.empty), VoteCalculatedState(Map.empty))
 
-      override def validateData(oldState: State, updates: NonEmptyList[Signed[PollUpdate]])(implicit context: L0NodeContext[IO]): IO[DataApplicationValidationErrorOr[Unit]] = MainData.validateData(oldState, updates)
+      override def validateUpdate(update: PollUpdate)(implicit context: L0NodeContext[IO]): IO[DataApplicationValidationErrorOr[Unit]] = IO {
+        ().validNec
+      }
```
sdavidp:
```suggestion
      override def validateUpdate(update: PollUpdate)(implicit context: L0NodeContext[IO]): IO[DataApplicationValidationErrorOr[Unit]] = ().validNec.pure[IO]
```

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/Main.scala
sdavidp, line 1: Similar comments as earlier re. calling shared code in `Main`

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/custom_routes/CustomRoutes.scala
```diff
@@ -0,0 +1,45 @@
...
+    if (poll.endSnapshotOrdinal < lastOrdinal) {
+      PollResponse(poll.id, poll.name, poll.owner, poll.pollOptions, poll.startSnapshotOrdinal, poll.endSnapshotOrdinal, "Closed")
+    } else {
+      PollResponse(poll.id, poll.name, poll.owner, poll.pollOptions, poll.startSnapshotOrdinal, poll.endSnapshotOrdinal, "Open")
+    }
+  }
+
+  def getAllPolls: IO[Response[IO]] = {
+    val calculatedState = getCalculatedState
+    calculatedState.flatMap { state =>
```
sdavidp: inline?

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/custom_routes/CustomRoutes.scala
```diff
@@ -0,0 +1,45 @@
...
+      val pollsResponse = state._2.polls.map { case (_, value) => formatPoll(value, state._1.value.value) }
+      Ok(pollsResponse)
+    }
+
+  }
+
+  def getPollById(pollId: String): IO[Response[IO]] = {
+    val calculatedState = getCalculatedState
+    calculatedState.flatMap { state =>
```
sdavidp: inline?

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,86 @@
...
+import org.tessellation.currency.dataApplication.{DataApplicationValidationError, DataState, L0NodeContext, L1NodeContext}
+import org.tessellation.security.signature.Signed
+import com.my.currency.shared_data.combiners.Combiners.{combineCreatePoll, combineVoteInPoll}
+import com.my.currency.shared_data.errors.Errors.CouldNotGetLatestCurrencySnapshot
+import com.my.currency.shared_data.validations.Validations.{createPollValidations, createPollValidationsWithSignature, voteInPollValidations, voteInPollValidationsWithSignature}
+import com.my.currency.shared_data.types.Types.{CreatePoll, PollUpdate, VoteCalculatedState, VoteInPoll, VoteStateOnChain}
+import org.slf4j.LoggerFactory
+import org.tessellation.security.SecurityProvider
+
+object Main {
```
sdavidp: If this file is not meant to be executed then maybe it shouldn't be called `Main`

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,86 @@
...
+
+  private val logger = LoggerFactory.getLogger("Main")
+
+  def validateUpdate(update: PollUpdate)(implicit context: L1NodeContext[IO]): IO[DataApplicationValidationErrorOr[Unit]] = {
```
sdavidp: Using `IO` in classes outside of the top-most class is discouraged.

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,86 @@
...
+  def validateData(state: DataState[VoteStateOnChain, VoteCalculatedState], updates: NonEmptyList[Signed[PollUpdate]])(implicit context: L0NodeContext[IO]): IO[DataApplicationValidationErrorOr[Unit]] = {
+    implicit val sp: SecurityProvider[IO] = context.securityProvider
+    updates.traverse { signedUpdate =>
+      signedUpdate.value match {
+        case poll: CreatePoll =>
+          createPollValidationsWithSignature(poll, signedUpdate.proofs, state)
+        case voteInPoll: VoteInPoll =>
+          getLastMetagraphIncrementalSnapshotInfo(Left(context)) match {
+            case Some(snapshotInfo) => voteInPollValidationsWithSignature(voteInPoll, signedUpdate.proofs, state, snapshotInfo)
+            case None => IO(CouldNotGetLatestCurrencySnapshot.asInstanceOf[DataApplicationValidationError].invalidNec)
```
sdavidp: 1. The result is a constant so it can be lifted into the effect with `pure`
2. `.asInstanceOf` ? Is this really necessary? If it is, that raises more concerns about the way the error class was defined.

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,86 @@
...
+      }
+    }.map(_.reduce)
+  }
+
+  def combine(state: DataState[VoteStateOnChain, VoteCalculatedState], updates: List[Signed[PollUpdate]])(implicit context: L0NodeContext[IO]): IO[DataState[VoteStateOnChain, VoteCalculatedState]] = {
+    val newStateIO = IO(DataState(VoteStateOnChain(List.empty), state.calculated))
+
+    if (updates.isEmpty) {
+      logger.info("Snapshot without any update, updating the state to empty updates")
+      return newStateIO
```
sdavidp: Use of `return` in Scala is discouraged as it breaks referential transparency.

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    getLastMetagraphIncrementalSnapshotInfo(Left(context)) match {
+      case None => println("Could not get lastMetagraphIncrementalSnapshotInfo, keeping current state")
+        IO(state)
```
sdavidp: use `pure`
Also, probably should use logger instead of `println`

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,86 @@
...
+          }
+          }
+          IO(updatedState)
+        })
+    }
+  }
+
+  def dataEncoder: Encoder[PollUpdate] = deriveEncoder
+
+  def dataDecoder: Decoder[PollUpdate] = deriveDecoder
```
sdavidp: Why are we deriving these? Aren't they already defined by the `derive` macros?

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Main.scala
```diff
@@ -0,0 +1,86 @@
...
+            val update = signedUpdate.value
+            update match {
+              case poll: CreatePoll =>
+                combineCreatePoll(poll, acc)
+              case voteInPoll: VoteInPoll =>
+                combineVoteInPoll(voteInPoll, acc, snapshotInfo)
+            }
+          }
+          }
+          IO(updatedState)
```
sdavidp: `pure`

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Utils.scala
```diff
@@ -1,33 +1,28 @@
...
 object Utils {
-  def customUpdateSerialization(update: PollUpdate): Array[Byte] = {
-    println("Serialize UPDATE event received")
-    println(update.asJson.deepDropNullValues.noSpaces)
-    update.asJson.deepDropNullValues.noSpaces.getBytes(StandardCharsets.UTF_8)
+  private def getSnapshotInfo(snapshotIO: IO[Option[(Hashed[CurrencyIncrementalSnapshot], CurrencySnapshotInfo)]]): Option[CurrencySnapshotInfo] = {
+    snapshotIO.map {
+      case Some(value) => Some(value._2)
+      case None => None
+    }.unsafeRunSync()
```
sdavidp: **!! Use of `unsafeRunSync` within the program is highly discouraged !!**

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/Utils.scala
```diff
@@ -1,33 +1,28 @@
...
-    parser.parse(new String(bytes, StandardCharsets.UTF_8)).flatMap { json =>
-      json.as[PollUpdate]
+  def getLastMetagraphIncrementalSnapshotInfo(context: Either[L0NodeContext[IO], L1NodeContext[IO]]): Option[CurrencySnapshotInfo] = {
+    context match {
+      case Left(value) =>
+        val lastSnapshotCombinedIO = value.getLastCurrencySnapshotCombined
+        getSnapshotInfo(lastSnapshotCombinedIO)
+      case Right(value) =>
+        val lastSnapshotCombinedIO = value.getLastCurrencySnapshotCombined
+        getSnapshotInfo(lastSnapshotCombinedIO)
```
sdavidp: for future, this might be written more concisely using `fold`

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/calculated_state/CalculatedState.scala
sdavidp, line 1: Several concerns with this file:

1. Can use `pure` for constant effects
2. Use of `IO` in non top-level class
3. MUTABLE SHARED STATE!  Are we not concerned that. multiple threads could be touching this value?
4. Leaked exceptions in calculating hash.

sdavidp (changes requested):
I haven't reviewed everything yet but there are some concerns that should be addressed.

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/validations/Validations.scala
```diff
@@ -0,0 +1,86 @@
...
+import org.tessellation.currency.dataApplication.DataState
+import org.tessellation.currency.dataApplication.dataApplication.DataApplicationValidationErrorOr
+import org.tessellation.currency.schema.currency.CurrencySnapshotInfo
+import org.tessellation.schema.SnapshotOrdinal
+import org.tessellation.security.SecurityProvider
+import org.tessellation.security.hash.Hash
+import org.tessellation.security.signature.signature.SignatureProof
+
+object Validations {
+  def createPollValidations[F[_] : Applicative](update: CreatePoll, maybeState: Option[DataState[VoteStateOnChain, VoteCalculatedState]], lastSnapshotOrdinal: Option[SnapshotOrdinal]): F[DataApplicationValidationErrorOr[Unit]] = {
```
sdavidp: This function may leak exceptions. Consider wrapping it in an `Async[F].delay`

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/validations/Validations.scala
```diff
@@ -0,0 +1,86 @@
...
+
+    }
+  }
+
+  def createPollValidationsWithSignature[F[_] : Async](update: CreatePoll, proofs: NonEmptySet[SignatureProof], state: DataState[VoteStateOnChain, VoteCalculatedState])(implicit sp: SecurityProvider[F]): F[DataApplicationValidationErrorOr[Unit]] = {
+    val validateAddressF = proofs
+      .map(_.id)
+      .toList
+      .traverse(_.toAddress[F])
+      .map(validateProvidedAddress(_, update.owner))
```
sdavidp:
```suggestion
    val validateAddressF = extractAddresses(proofs)
```
where
```
private def extractAddresses[F[_] : Async : SecurityProvider](proofs: NonEmptySet[SignatureProof]) = {
  proofs
    .map(_.id)
    .toList
    .traverse(_.toAddress[F])
}
```

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/validations/Validations.scala
```diff
@@ -0,0 +1,86 @@
...
+      .toList
+      .traverse(_.toAddress[F])
+      .map(validateProvidedAddress(_, update.address))
+
+    val validations = voteInPollValidations(update, state.some, None, snapshotInfo)
+
+    for {
+      validatedAddress <- validateAddress
+      validatedPoll <- validations
+    } yield validatedAddress.productR(validatedPoll)
```
sdavidp:
```suggestion
    for {
      addresses <- extractAddresses(proofs)
      validatedAddress = validateProvidedAddress(addresses, update.address)
      validatedPoll <- voteInPollValidations(update, state.some, None, snapshotInfo)
    } yield validatedAddress.productR(validatedPoll)
```

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/validations/TypeValidators.scala
sdavidp, line 1: If you decide to use the `Ops` class suggested in `Errors` you can simplify this module to the following. (Be sure to verify I've translated the predicates correctly.)
```
  def validateIfPollAlreadyExists(state: DataState[VoteStateOnChain, VoteCalculatedState], pollId: String): DataApplicationValidationType =
    PollAlreadyExists.whenA(state.calculated.polls.contains(pollId))

  def validateIfVotePollExists(state: DataState[VoteStateOnChain, VoteCalculatedState], pollId: String): DataApplicationValidationType =
    PollDoesNotExists.unlessA(state.calculated.polls.contains(pollId))

  def validateIfOptionExists(state: DataState[VoteStateOnChain, VoteCalculatedState], pollId: String, option: String): DataApplicationValidationType =
    InvalidOption.whenA(state.calculated.polls.get(pollId).exists(_.pollOptions.contains(option)))

  def validateProvidedAddress(proofAddresses: List[Address], address: Address): DataApplicationValidationType =
    InvalidAddress.unlessA(proofAddresses.contains(address))

  def validateIfUserAlreadyVoted(state: DataState[VoteStateOnChain, VoteCalculatedState], pollId: String, address: Address): DataApplicationValidationType =
    RepeatedVote.whenA(state.calculated.polls.get(pollId).map(_.usersVotes).exists(_.contains(address)))

  def validateWalletBalance(snapshotInfo: CurrencySnapshotInfo, walletAddress: Address): DataApplicationValidationType =
    NotEnoughWalletBalance.unlessA(snapshotInfo.balances.get(walletAddress).exists(_.value.value > 0L))

  def validateSnapshotCreatePoll(snapshotOrdinal: SnapshotOrdinal, update: CreatePoll): DataApplicationValidationType =
    InvalidEndSnapshot.whenA(update.endSnapshotOrdinal < snapshotOrdinal.value.value)

  def validatePollSnapshotInterval(lastSnapshotOrdinal: SnapshotOrdinal, state: DataState[VoteStateOnChain, VoteCalculatedState], voteInPoll: VoteInPoll): DataApplicationValidationType =
    state.calculated.polls
      .get(voteInPoll.pollId)
      .map(value => ClosedPool.whenA(value.endSnapshotOrdinal < lastSnapshotOrdinal.value.value))
      .getOrElse(PollDoesNotExists.invalid)
```

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/custom_routes/CustomRoutes.scala
```diff
@@ -0,0 +1,52 @@
...
+    getCalculatedState.flatMap { state =>
+      val poll = state._2.polls.get(pollId)
+      poll match {
+        case Some(value) => Ok(formatPoll(value, state._1.value.value))
+        case None => NotFound()
+      }
+    }
+  }
+
+   val public: HttpRoutes[F] = HttpRoutes.of[F] {
```
sdavidp: Add the `CORS.policy. . . .` here?

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/custom_routes/CustomRoutes.scala
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }
+
+  private def getAllPolls: F[Response[F]] = {
+    getCalculatedState.flatMap { state =>
+      val pollsResponse = state._2.polls.map { case (_, value) => formatPoll(value, state._1.value.value) }
+      Ok(pollsResponse)
+    }
+
+  }
```
sdavidp: Is this correct? What are we returning if `getCalculatedState` or the composed function raises an error?

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/LifecycleSharedFunctions.scala
```diff
@@ -0,0 +1,82 @@
...
+      update match {
+        case poll: CreatePoll =>
+          createPollValidations(poll, None, Some(lastSnapshotOrdinal))
+        case voteInPoll: VoteInPoll =>
+          getLastMetagraphIncrementalSnapshotInfo(context.asRight[L0NodeContext[F]]).flatMap {
+            case Some(snapshotInfo) => voteInPollValidations(voteInPoll, None, Some(lastSnapshotOrdinal), snapshotInfo)
+            case None => CouldNotGetLatestCurrencySnapshot.asInstanceOf[DataApplicationValidationError].invalidNec[Unit].pure[F]
+          }
+      }
+    }
```
sdavidp:
```suggestion
    context
      .getLastCurrencySnapshot
      .map(_.get.ordinal)
      .flatMap { lastSnapshotOrdinal =>
        update match {
          case poll: CreatePoll => createPollValidations(poll, none, lastSnapshotOrdinal.some)

          case voteInPoll: VoteInPoll =>
            getLastMetagraphIncrementalSnapshotInfo(context.asRight[L0NodeContext[F]])
              .flatMap {
                case Some(snapshotInfo) => voteInPollValidations(voteInPoll, none, lastSnapshotOrdinal.some, snapshotInfo)
                case None => CouldNotGetLatestCurrencySnapshot.invalidNec[Unit].pure[F]
              }
        }
      }
```
You'll need `import cats.syntax.option._` for the `none` and `.some` -- it's a good habit to use these since in certain circumstances they help the compiler with type inference.

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/LifecycleSharedFunctions.scala
```diff
@@ -0,0 +1,82 @@
+package com.my.currency.shared_data
+
+import cats.Monad
+import cats.conversions.all.autoWidenFunctor
+import cats.data.NonEmptyList
+import cats.effect.Async
+import cats.implicits.{catsSyntaxApplicativeId, catsSyntaxEitherId, catsSyntaxValidatedIdBinCompat0, toFlatMapOps, toFunctorOps}
```
sdavidp:
```suggestion
import cats.syntax.applicative._
import cats.syntax.either._
import cats.syntax.flatMap._
import cats.syntax.functor._
import cats.syntax.option._
import cats.syntax.validated._
```

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/errors/Errors.scala
sdavidp, line 1: Consider adding this to the `Errors` object. It can help simplify a lot of the validation functions.
```
type DataApplicationValidationType = DataApplicationValidationErrorOr[Unit]

val valid: DataApplicationValidationType = ().validNec[DataApplicationValidationError]

implicit class DataApplicationValidationTypeOps[E <: DataApplicationValidationError](err: E) {
  def invalid: DataApplicationValidationType = err.invalidNec[Unit]
  def unlessA(cond: Boolean): DataApplicationValidationType = if (cond) valid else invalid
  def whenA(cond: Boolean):   DataApplicationValidationType = if (cond) invalid else valid
}
```

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/calculated_state/CalculatedState.scala
sdavidp, line 1: This file has been discussed in other places. The only comment I have is once we update to use the revised `dataApplication` hook we'll replace the `AtomicReference` with a `Ref`

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/types/Types.scala
```diff
@@ -0,0 +1,36 @@
...
+    val name: String
+    val owner: Address
+    val pollOptions: Map[String, Long]
+    val usersVotes: Map[Address, Map[String, Long]]
+    val startSnapshotOrdinal: Long
+    val endSnapshotOrdinal: Long
+  }
+
+  @derive(decoder, encoder)
+  case class Poll(id: String, name: String, owner: Address, pollOptions: Map[String, Long], usersVotes: Map[Address, Map[String, Long]], startSnapshotOrdinal: Long, endSnapshotOrdinal: Long) extends PollState
```
sdavidp: This is the only implementation of the trait. Is there any reason why we aren't just calling the class `PollState` ?

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/validations/Validations.scala
```diff
@@ -0,0 +1,86 @@
...
+          case None => ().validNec
+        }
+
+        val validatedPoll = validateIfVotePollExists(state, update.pollId)
+
+        val validatedOption = validateIfOptionExists(state, update.pollId, update.option)
+
+        val validatedRepeatedVote = validateIfUserAlreadyVoted(state, update.pollId, update.address)
+
+        validatedSnapshotInterval.productR(validatedPoll).productR(validatedOption).productR(validatedRepeatedVote).productR(validateBalance).pure
```
sdavidp:
```suggestion
        validatedSnapshotInterval
          .productR(validatedPoll)
          .productR(validatedOption)
          .productR(validatedRepeatedVote)
          .productR(validateBalance)
          .pure
```
It may be easier to read with newlines

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/validations/Validations.scala
```diff
@@ -0,0 +1,86 @@
...
+import org.tessellation.schema.SnapshotOrdinal
+import org.tessellation.security.SecurityProvider
+import org.tessellation.security.hash.Hash
+import org.tessellation.security.signature.signature.SignatureProof
+
+object Validations {
+  def createPollValidations[F[_] : Applicative](update: CreatePoll, maybeState: Option[DataState[VoteStateOnChain, VoteCalculatedState]], lastSnapshotOrdinal: Option[SnapshotOrdinal]): F[DataApplicationValidationErrorOr[Unit]] = {
+    val validatedCreatePollSnapshot = lastSnapshotOrdinal match {
+      case Some(value) => validateSnapshotCreatePoll(value, update)
+      case None => ().validNec
```
sdavidp:
```suggestion
      case None => valid
```

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/serializers/Serializers.scala
sdavidp, line 1: Do we really need to do those `println`s ? Even if we use a logger won't that be very noisy?
Also, be aware that `.getBytes` is a Java function that can throw. I've raised this issue in the past and some believe it will not throw due to how we're using it. I'll leave it up to the author to decide what - if anything - to do about it.

Lastly, all three functions are doing exactly the same thing. Couldn't we replace all three with this?
```
def serialize[A: Encoder](state: A): Array[Byte] =
  state.asJson.deepDropNullValues.noSpaces.getBytes(StandardCharsets.UTF_8)
```

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/deserializers/Deserializers.scala
```diff
@@ -0,0 +1,30 @@
...
+    parser.parse(new String(bytes, StandardCharsets.UTF_8)).flatMap { json =>
+      json.as[VoteStateOnChain]
+    }
+  }
+
+  def deserializeBlock(bytes: Array[Byte])(implicit e: Decoder[DataUpdate]): Either[Throwable, Signed[DataApplicationBlock]] = {
+    parser.parse(new String(bytes, StandardCharsets.UTF_8)).flatMap { json =>
+      json.as[Signed[DataApplicationBlock]]
+    }
+  }
```
sdavidp:
```suggestion
  def deserialize[A: Decoder](bytes: Array[Byte]): Either[circe.Error, A] =
    parser
      .parse(new String(bytes, StandardCharsets.UTF_8))
      .flatMap(_.as[A])
```

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/Main.scala
```diff
@@ -2,55 +2,75 @@ package com.my.currency.l0
...
+    override def serializeBlock(block: Signed[DataApplicationBlock]): IO[Array[Byte]] = IO(Serializers.serializeBlock(block)(dataEncoder.asInstanceOf[Encoder[DataUpdate]]))
 
-      override def deserializeUpdate(bytes: Array[Byte]): IO[Either[Throwable, PollUpdate]] = MainData.deserializeUpdate(bytes)
+    override def deserializeState(bytes: Array[Byte]): IO[Either[Throwable, VoteStateOnChain]] = IO(Deserializers.deserializeState(bytes))
 
-      override def dataEncoder: Encoder[PollUpdate] = MainData.dataEncoder
+    override def deserializeUpdate(bytes: Array[Byte]): IO[Either[Throwable, PollUpdate]] = IO(Deserializers.deserializeUpdate(bytes))
 
-      override def dataDecoder: Decoder[PollUpdate] = MainData.dataDecoder
+    override def deserializeBlock(bytes: Array[Byte]): IO[Either[Throwable, Signed[DataApplicationBlock]]] = IO(Deserializers.deserializeBlock(bytes)(dataDecoder.asInstanceOf[Decoder[DataUpdate]]))
```
sdavidp, line 56:
```suggestion
    override def serializeState(state: VoteStateOnChain): IO[Array[Byte]] =
      IO(Serializers.serialize[VoteStateOnChain](state))

    override def serializeUpdate(update: PollUpdate): IO[Array[Byte]] =
      IO(Serializers.serialize[PollUpdate](update))

    override def serializeBlock(block: Signed[DataApplicationBlock]): IO[Array[Byte]] =
      IO {
        implicit val dataUpdateEncoder = dataEncoder.asInstanceOf[Encoder[DataUpdate]]
        Serializers.serialize[DataApplicationBlock](block)
      }

    override def deserializeState(bytes: Array[Byte]): IO[Either[Throwable, VoteStateOnChain]] =
      IO(Deserializers.deserialize[VoteStateOnChain](bytes))

    override def deserializeUpdate(bytes: Array[Byte]): IO[Either[Throwable, PollUpdate]] =
      IO(Deserializers.deserialize[PollUpdate](bytes))

    override def deserializeBlock(bytes: Array[Byte]): IO[Either[Throwable, Signed[DataApplicationBlock]]] =
      IO {
        implicit val dataUpdateDecoder: Decoder[DataUpdate] = dataDecoder.asInstanceOf[Decoder[DataUpdate]]
        Deserializers.deserialize[Signed[DataApplicationBlock]](bytes)
      }
```

examples/DataApi-Voting-poll/template/modules/data_l1/src/main/scala/com/my/currency/data_l1/Main.scala
sdavidp, line 1: See comments in `l0/Main` for suggested changes to the serialize/deserialize methods.

sdavidp:
> I haven't reviewed everything yet but there are some concerns that should be addressed.

Just spent time reviewing. Let me know if you have any questions/concerns.

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/validations/Validations.scala
marcus-girolneto: Updated

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/validations/Validations.scala
marcus-girolneto: Updated

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/validations/Validations.scala
marcus-girolneto: Updated

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/validations/TypeValidators.scala
marcus-girolneto: Thanks, updated

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/custom_routes/CustomRoutes.scala
marcus-girolneto: Resolved

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/custom_routes/CustomRoutes.scala
marcus-girolneto: In this case, it doesn't throw errors, we only return what we have in memory. We always will have something in memory

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/LifecycleSharedFunctions.scala
marcus-girolneto: Got it updated

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/LifecycleSharedFunctions.scala
marcus-girolneto: Updated

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/errors/Errors.scala
marcus-girolneto: Added

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/calculated_state/CalculatedState.scala
marcus-girolneto: Okay, we can update when the version is released

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/types/Types.scala
marcus-girolneto: We can remove

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/validations/Validations.scala
marcus-girolneto: Yeah, updated

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/validations/Validations.scala
marcus-girolneto: Fixed

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/serializers/Serializers.scala
marcus-girolneto: Hmm good catch, I've updated to use your suggestion

marcus-girolneto: I prefer keep with names and call the shared function

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/deserializers/Deserializers.scala
marcus-girolneto: I prefer to keep with names and call the shared function serialize/deserialize

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/Main.scala
marcus-girolneto: I prefer to keep with names and call the shared function serialize/deserialize

examples/DataApi-Voting-poll/template/modules/data_l1/src/main/scala/com/my/currency/data_l1/Main.scala
marcus-girolneto: I prefer to keep with names and call the shared function serialize/deserialize

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/calculated_state/CalculatedState.scala
```diff
@@ -0,0 +1,51 @@
...
+import io.circe.syntax.EncoderOps
+import org.tessellation.schema.SnapshotOrdinal
+import org.tessellation.security.hash.Hash
+
+import java.security.MessageDigest
+import java.util.concurrent.atomic.AtomicReference
+
+object CalculatedState {
+
+  private val maybeVoteCalculatedState: AtomicReference[(SnapshotOrdinal, VoteCalculatedState)] = new AtomicReference(
```
sdavidp: BTW, I missed this before:
The `maybe` prefix is commonly used to signify an `Option`, which this is not. If it were me I'd simply call this `stateRef`.

marcus-girolneto: Got it, I'll update

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/custom_routes/CustomRoutes.scala
```diff
@@ -43,10 +44,16 @@ case class CustomRoutes[F[_]: Async]() extends Http4sDsl[F] with PublicRoutes[F]
...
     case GET -> Root / "polls" => getAllPolls
     case GET -> Root / "polls" / poolId => getPollById(poolId)
   }
 
+  val public: HttpRoutes[F] =
+    CORS
+      .policy
+      .withAllowCredentials(true)
+      .httpRoutes(routes)
+
```
sdavidp: This needs to be confirmed by someone else.
@AlexBrandes and/or @scasplte2 maybe?
What I provided was sample usage -- I don't know what the actual settings need to be.

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/LifecycleSharedFunctions.scala
```diff
@@ -20,20 +24,21 @@ object LifecycleSharedFunctions {
...
+      .getLastCurrencySnapshot
+      .map(_.get.ordinal)
+      .flatMap { lastSnapshotOrdinal =>
+        update match {
+          case poll: CreatePoll => createPollValidations(poll, none, lastSnapshotOrdinal.some)
+          case voteInPoll: VoteInPoll =>
+            getLastMetagraphIncrementalSnapshotInfo(context.asRight[L0NodeContext[F]])
+              .flatMap {
+                case Some(snapshotInfo) => voteInPollValidations(voteInPoll, none, lastSnapshotOrdinal.some, snapshotInfo)
+                case None => CouldNotGetLatestCurrencySnapshot.invalidNec[Unit].pure[F]
```
sdavidp:
```suggestion
                case None => CouldNotGetLatestCurrencySnapshot.invalid.pure[F]
```

sdavidp (approved):
Left a couple suggestions otherwise LGTM

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/Main.scala
```diff
@@ -2,55 +2,75 @@ package com.my.currency.l0
 
 import cats.data.NonEmptyList
 import cats.effect.IO
-import cats.implicits.catsSyntaxValidatedIdBinCompat0
-import com.my.currency.shared_data.MainData
-import com.my.currency.shared_data.MainData.{PollUpdate, State}
+import cats.implicits.{catsSyntaxApplicativeId, catsSyntaxValidatedIdBinCompat0}
```
sdavidp:
```suggestion
import cats.syntax.applicative._
import cats.syntax.validated._
```

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/custom_routes/CustomRoutes.scala
sdavidp: But that may change one day, such as when the implementation of `getCalculatedState` changes to using a `Ref`?

The point is, what is the response when an error is raised in `F` ?

examples/DataApi-Voting-poll/template/modules/data_l1/src/main/scala/com/my/currency/data_l1/Main.scala
```diff
@@ -2,20 +2,23 @@ package com.my.currency.data_l1
 
 import cats.data.NonEmptyList
 import cats.effect.IO
-import cats.implicits.catsSyntaxValidatedIdBinCompat0
-import com.my.currency.data_l1.CustomRoutes.{getAllPolls, getPollById}
-import com.my.currency.shared_data.MainData
-import com.my.currency.shared_data.MainData.{PollUpdate, State}
+import cats.implicits.{catsSyntaxApplicativeId, catsSyntaxValidatedIdBinCompat0}
```
sdavidp:
```suggestion
import cats.syntax.applicative._
import cats.syntax.validated._
```

examples/DataApi-Voting-poll/template/modules/shared_data/src/main/scala/com/my/currency/shared_data/LifecycleSharedFunctions.scala
marcus-girolneto: Done

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/Main.scala
marcus-girolneto: Done

examples/DataApi-Voting-poll/template/modules/data_l1/src/main/scala/com/my/currency/data_l1/Main.scala
marcus-girolneto: Done

marcus-girolneto: @sdavidp about the point of getCalculatedState, I think we may want to return a 400 or 500 and with some message that state failed to be obtained

sdavidp: Are you sure? That should never fail so it makes sense to me that if that happens we let the error bubble up and let the high-level handler (already provided for us) report the error.

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/custom_routes/CustomRoutes.scala
```diff
@@ -0,0 +1,66 @@
+package com.my.currency.l0.custom_routes
+
+import cats.effect.Async
+import cats.implicits.{catsSyntaxApplicativeError, catsSyntaxFlatMapOps, toFlatMapOps}
```
sdavidp:
```suggestion
import cats.syntax.applicativeError._
import cats.syntax.flatMap._
import cats.syntax.functor._
```

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/custom_routes/CustomRoutes.scala
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+
+  private def getAllPolls: F[Response[F]] = {
+    getCalculatedState
+      .flatMap { state =>
+        val pollsResponse = state._2.polls.map { case (_, value) => formatPoll(value, state._1.value.value) }
+        Ok(pollsResponse)
+      }
+      .handleErrorWith(e => logger.error(e)(s"An error occured!") >> BadRequest())
```
sdavidp:
```suggestion
      .map { case (ord, state) => state.polls.view.mapValues(formatPoll(_, ord.value.value)).toList }
      .flatMap(Ok(_))
      .handleErrorWith(e => logger.error(e)("Error getting all polls") >> BadRequest())
```
What do you think of this refactor? If you think it reads better then go for it.

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/custom_routes/CustomRoutes.scala
```diff
@@ -0,0 +1,66 @@
...
+  private def getPollById(pollId: String): F[Response[F]] = {
+    getCalculatedState
+      .flatMap { state =>
+        val poll = state._2.polls.get(pollId)
+        poll match {
+          case Some(value) => Ok(formatPoll(value, state._1.value.value))
+          case None => NotFound()
+        }
+      }
+      .handleErrorWith(e => logger.error(e)(s"An error occured!") >> BadRequest())
```
sdavidp:
```suggestion
      .map { case (ord, state) => state.polls.get(pollId).map(formatPoll(_, ord.value.value)) }
      .flatMap(_.fold(NotFound())(Ok(_)))
      .handleErrorWith(e => logger.error(e)(s"Error getting poll for id $pollId") >> BadRequest())
```
Same here. If you like this refactor then use it.

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/custom_routes/CustomRoutes.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  private val routes: HttpRoutes[F] = HttpRoutes.of[F] {
+    case GET -> Root / "polls" => getAllPolls
+    case GET -> Root / "polls" / poolId => getPollById(poolId)
+  }
+
+  val public: HttpRoutes[F] =
+    CORS
+      .policy
+      .withAllowCredentials(false)
```
sdavidp, line 64: don't forget to check (with Alex?) to confirm what the CORS needs to be set to.

examples/DataApi-Voting-poll/template/modules/data_l1/src/main/scala/com/my/currency/data_l1/Main.scala
marcus-girolneto: Got it, fixed

examples/DataApi-Voting-poll/template/modules/l0/src/main/scala/com/my/currency/l0/custom_routes/CustomRoutes.scala
marcus-girolneto: I think these ones are fine for now. @AlexBrandes can confirm, but the customer that sees the example can change according to their metagraph

marcus-girolneto:
@sdavidp I've applied your suggestions and merged the PR

sdavidp:
👍

PR #9, opened 2023-11-17
Updating NFT Example

marcus-girolneto:
* Updating NFT example to use calculatedState


examples/DataApi-NFT/README.md
```diff
@@ -62,99 +62,119 @@ The NFT management system operates as follows:
...
+- `GET <metagraph l0 url>/data-application/collections`: Retrieves all collections.
+- `GET <metagraph l0 url>/data-application/collections/:collection_id`: Retrieves a collection by ID.
+- `GET <metagraph l0 url>/data-application/collections/:collection_id/nfts`: Retrieves all NFTs of a collection.
+- `GET <metagraph l0 url>/data-application/collections/:collection_id/nfts/:nft_id`: Retrieves an NFT of a collection by ID.
+- `GET <metagraph l0 url>/data-application/addresses/:address/collections`: Retrieves all collections associated with an address.
+- `GET <metagraph l0 url>/data-application/addresses/:address/nfts`: Retrieves all NFTs associated with an address.
+
+## Scripts
+
+The example includes a script named `send_data_transaction.js` for generating, signing, and sending data updates to the Metagraph. To use the script:
```
AlexBrandes: I think we should update the script to take inputs rather than require the developer to edit the file directly. In talking with developers using these examples, this part seems to add confusion. 

We could also reference the scripts in the UI app if that's easier.

examples/DataApi-NFT/template/modules/l0/src/main/scala/com/my/nft_example/l0/Main.scala
```diff
@@ -26,32 +37,54 @@ object Main
...
-      override def routes(implicit context: L0NodeContext[IO]): HttpRoutes[IO] = HttpRoutes.empty
+      override def dataEncoder: Encoder[NFTUpdate] = implicitly[Encoder[NFTUpdate]]
+
+      override def dataDecoder: Decoder[NFTUpdate] = implicitly[Decoder[NFTUpdate]]
+
+      override def calculatedStateEncoder: Encoder[NFTUpdatesCalculatedState] = implicitly[Encoder[NFTUpdatesCalculatedState]]
+
+      override def calculatedStateDecoder: Decoder[NFTUpdatesCalculatedState] = implicitly[Decoder[NFTUpdatesCalculatedState]]
+
+      override def routes(implicit context: L0NodeContext[IO]): HttpRoutes[IO] = HttpRoutes.of {
```
AlexBrandes: Are we able to add CORS headers to these responses? We've had issues where people are not able to access the endpoints from the frontend app because they're missing an allow-all CORS header.

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/calculated_state/CalculatedState.scala
```diff
@@ -0,0 +1,49 @@
...
+
+    maybeNFTUpdatesCalculatedState = (
+      snapshotOrdinal,
+      NFTUpdatesCalculatedState(updatedDevices)
+    )
+
+    IO(true)
+  }
+
+  private def sha256Hash(input: String): String = {
```
scasplte2: this is private but doesn't seem used here. If you do need it though, perhaps the sdk methods in `org.tessellation.security.hash.Hash` would be better suited?

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/calculated_state/CalculatedState.scala
```diff
@@ -0,0 +1,49 @@
...
+  private var maybeNFTUpdatesCalculatedState: (SnapshotOrdinal, NFTUpdatesCalculatedState) = (
+    SnapshotOrdinal(NonNegLong(0L)),
+    NFTUpdatesCalculatedState(Map.empty)
+  )
+
+  def getCalculatedState: IO[(SnapshotOrdinal, NFTUpdatesCalculatedState)] = {
+    IO(maybeNFTUpdatesCalculatedState)
+  }
+
+  def setCalculatedState(snapshotOrdinal: SnapshotOrdinal, state: NFTUpdatesCalculatedState): IO[Boolean] = {
```
scasplte2: It would be preferable to move `IO` out of this layer of that application but if we keep it here, I think we should at least wrap the entire computation in the `IO` with `IO.delay { ... }`

examples/DataApi-NFT/template/modules/l0/src/main/scala/com/my/nft_example/l0/Main.scala
scasplte2: @IPadawans I believe I have a way to do this in flight data at https://github.com/Constellation-Labs/flight-data/blob/main/source/project/flight-data/modules/data_l1/src/main/scala/com/my/nft_example/data_l1/CustomRoutes.scala

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/calculated_state/CalculatedState.scala
```diff
@@ -0,0 +1,49 @@
...
+import eu.timepit.refined.types.numeric.NonNegLong
+import io.circe.syntax.EncoderOps
+import org.tessellation.schema.SnapshotOrdinal
+import org.tessellation.security.hash.Hash
+
+import java.security.MessageDigest
+
+object CalculatedState {
+
+  private var maybeNFTUpdatesCalculatedState: (SnapshotOrdinal, NFTUpdatesCalculatedState) = (
```
scasplte2: @AlexBrandes this is the internal variable that I mentioned yesterday. The patch Krystian is working on doesn't apply in this case because there needs to be a place to store the `NFTCalculatedState`. This would likely need to move to some on-disk persistence solution but perhaps this has been a discussion with @ryle-goehausen and the wider protocol team already.

scasplte2 (commented):
I think there needs to be a discussion about the internal class variable within `CalculatedState` but otherwise looks reasonable.

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/calculated_state/CalculatedState.scala
sdavidp: As it is this is definitely not safe and has the potential for race conditions on other threads. Yesterday I showed @IPadawans how to wrap it in an `AtomicReference` but now that the _dataApplication_ hook has been updated we can use the `make` pattern to create our services and use a `Ref`.

examples/DataApi-NFT/template/modules/l0/src/main/scala/com/my/nft_example/l0/Main.scala
sdavidp: @IPadawans @AlexBrandes @scasplte2 
Maybe we should use the code already available to us in _http4s_, as we're doing in Tessellation? I don't know what settings are required but maybe we should start with something like this:
```
val public: HttpRoutes[F] =
  CORS
    .policy
    .withAllowCredentials(true)
    .httpRoutes(routes)
```

sdavidp: Addendum to above, where in this example `routes` is defined as:
```
 private val routes: HttpRoutes[F] = HttpRoutes.of[F] {
  case GET -> Root / "polls" => getAllPolls
  case GET -> Root / "polls" / poolId => getPollById(poolId)
}
```

sdavidp: In tessellation they're doing this:
```
CORS.policy.withAllowOriginAll.withAllowHeadersAll.withAllowCredentials(false) . . .
```

marcus-girolneto: Added

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/calculated_state/CalculatedState.scala
marcus-girolneto: Good catch

examples/DataApi-NFT/README.md
marcus-girolneto: I think we can reference, because the scripts already exists

examples/DataApi-NFT/template/modules/data_l1/src/main/scala/com/my/nft_example/data_l1/Main.scala
```diff
@@ -29,44 +32,106 @@ object Main
...
+    ): IO[Either[Throwable, NFTUpdatesCalculatedState]] =
+      IO(Deserializers.deserializeCalculatedState(bytes))
+  })
+
+  private def makeL1Service: IO[BaseDataApplicationL1Service[IO]] = {
+    for {
+      calculatedStateService <- CalculatedStateService.make[IO]
+      dataApplicationL1Service = makeBaseDataApplicationL1Service(calculatedStateService)
+    } yield dataApplicationL1Service
+  }
```
sdavidp:
```suggestion
  private def makeL1Service: IO[BaseDataApplicationL1Service[IO]] =
    CalculatedStateService.make[IO].map(makeBaseDataApplicationL1Service)
```

examples/DataApi-NFT/template/modules/l0/src/main/scala/com/my/nft_example/l0/Main.scala
```diff
@@ -36,55 +37,111 @@ object Main
...
+    ): IO[Either[Throwable, NFTUpdatesCalculatedState]] =
+      IO(Deserializers.deserializeCalculatedState(bytes))
+  })
+
+  private def makeL0Service: IO[BaseDataApplicationL0Service[IO]] = {
+    for {
+      calculatedStateService <- CalculatedStateService.make[IO]
+      dataApplicationL0Service = makeBaseDataApplicationL0Service(calculatedStateService)
+    } yield dataApplicationL0Service
+  }
```
sdavidp: same comment as earlier -- can be simplified to one line

examples/DataApi-NFT/template/modules/l0/src/main/scala/com/my/nft_example/l0/custom_routes/CustomRoutes.scala
```diff
@@ -1,68 +1,92 @@
...
+      nft.metadata
+    )
   }
 
-  private def getState: IO[NFTUpdatesCalculatedState] = {
-    val calculatedState = getCalculatedState
-    calculatedState.map(_._2)
+  private def getState: F[NFTUpdatesCalculatedState] = {
+    val calculatedState = calculatedStateService.getCalculatedState
+    calculatedState.map(_.state)
```
sdavidp: inline that val ?

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/Utils.scala
```diff
@@ -1,19 +1,43 @@
...
+import org.tessellation.security.SecurityProvider
+import org.tessellation.security.signature.signature.SignatureProof
+
 import java.net.{MalformedURLException, URISyntaxException, URL}
+import scala.util.{Failure, Success, Try}
 
 object Utils {
 
   @throws[MalformedURLException]
   @throws[URISyntaxException]
```
sdavidp: Is this leftover code that can be removed?

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/Utils.scala
```diff
@@ -1,19 +1,43 @@
...
-  } catch {
-    case _: MalformedURLException =>
-      false
-    case _: URISyntaxException =>
-      false
+  def isValidURL(url: String): Boolean =
+    Try(new URL(url).toURI) match {
+      case Failure(_) => false
+      case Success(_) => true
+    }
```
sdavidp:
```suggestion
    Try(new URL(url).toURI).isSuccess
```

I don't like exception-based flow control but I'm not aware of a better way to do this.

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/Utils.scala
```diff
@@ -1,19 +1,43 @@
...
+  }
+
+  def getFirstAddressFromProofs[F[_] : Async : SecurityProvider](
+    proofs: NonEmptySet[SignatureProof]
+  ): F[Address] = {
+    proofs
+      .map(_.id)
+      .toList
+      .head
+      .toAddress[F]
```
sdavidp:
```suggestion
    proofs.head.id.toAddress[F]
```
No need to construct a new Set and then a List if we only want the first element anyway

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/combiners/Combiners.scala
```diff
@@ -1,14 +1,18 @@
...
 
 object Combiners {
-  def combineMintCollection(update: MintCollection, state: DataState[NFTUpdatesState, NFTUpdatesCalculatedState], collectionOwner: Address): DataState[NFTUpdatesState, NFTUpdatesCalculatedState] = {
+  def combineMintCollection(
+    update         : MintCollection,
+    state          : DataState[NFTUpdatesState, NFTUpdatesCalculatedState],
+    collectionOwner: Address
+  ): DataState[NFTUpdatesState, NFTUpdatesCalculatedState] = {
     val collectionId = Hash.fromBytes(Serializers.serializeUpdate(update)).toString
     val nowInTime = System.currentTimeMillis()
```
sdavidp, line 17: For future reference, getting time this way makes it very hard to write unit tests. One technique in handling this is to have a date/time provider injected into your service. In Java we would create such an object but with Cats you can probably use `Clock` or something related.

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/deserializers/Deserializers.scala
```diff
@@ -9,22 +9,30 @@ import org.tessellation.security.signature.Signed
...
-
-  def deserializeUpdate(bytes: Array[Byte]): Either[Throwable, NFTUpdate] = {
+  private def deserialize[A: Decoder](
+    bytes: Array[Byte]
+  ): Either[Throwable, A] =
     parser.parse(new String(bytes, StandardCharsets.UTF_8)).flatMap { json =>
-      json.as[NFTUpdate]
+      json.as[A]
     }
-  }
```
sdavidp: I just realized. Aren't we effectively doing this?
```
decode[A](new String(bytes, StandardCharsets.UTF_8))
```

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/validations/Validations.scala
```diff
@@ -1,98 +1,128 @@
...
+          val validateCollection = validateIfCollectionIsUnique(update, state)
+          val validateCollectionName = validateStringMaxSize(update.name, 64, "name")
           validateCollection.productR(validateCollectionName)
         }
       case None =>
-        val validateCollectionName = validateStringMaxSize(update.name, 64, "name")
-        IO {
-          validateCollectionName
+        Async[F].delay {
+          validateStringMaxSize(update.name, 64, "name")
```
sdavidp: Since you're doing the `.delay` in both cases why not put the whole method in a single `.delay` ?

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/validations/Validations.scala
```diff
@@ -1,98 +1,128 @@
...
-        val validateNFTUrlValid = validateIfNFTUriIsValid(update)
-        val validateUniqueNFTURI = validateIfNFTUriIsUnique(update, state)
-        val validateUniqueNFTId = validateIfNFTIdIsUnique(update, state)
-        val validateNFTName = validateStringMaxSize(update.name, 64, "name")
-        val validateNFTDescription = validateStringMaxSize(update.description, 64, "description")
-        val validateNFTMetadata = validateMapMaxSize(update.metadata, 15, "metadata")
-
-        IO {
-          validateNFTUrlValid.productR(validateUniqueNFTURI).productR(validateUniqueNFTId).productR(validateNFTName).productR(validateNFTDescription).productR(validateNFTMetadata)
+        Async[F].delay {
```
sdavidp: Same comment as above

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/validations/Validations.scala
```diff
@@ -1,98 +1,128 @@
...
-  }
 
-  def mintNFTValidationsWithSignature(update: MintNFT, addresses: List[Address], state: DataState[NFTUpdatesState, NFTUpdatesCalculatedState]): IO[DataApplicationValidationErrorOr[Unit]] = {
-    val validateAddress = validateProvidedAddress(addresses, update.owner).pure[IO]
+  def mintNFTValidationsWithSignature[F[_] : Async](
+    update   : MintNFT,
+    addresses: List[Address],
+    state    : DataState[NFTUpdatesState, NFTUpdatesCalculatedState]
+  ): F[DataApplicationValidationErrorOr[Unit]] = {
+    val validateAddress = Async[F].delay(validateProvidedAddress(addresses, update.owner))
```
sdavidp: Hold on -- am I correct that in some places in this module you're doing a `productR` on `Validation` objects, but in this case you're doing it on `F` objects?

If so, I just checking to see if the latter case is going to be doing what you want.

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/validations/Validations.scala
```diff
@@ -1,98 +1,128 @@
...
   }
 
-  def transferCollectionValidationsWithSignature(update: TransferCollection, addresses: List[Address], state: DataState[NFTUpdatesState, NFTUpdatesCalculatedState]): IO[DataApplicationValidationErrorOr[Unit]] = {
-    val validateAddress = validateProvidedAddress(addresses, update.fromAddress).pure[IO]
+  def transferCollectionValidationsWithSignature[F[_] : Async](
+    update   : TransferCollection,
+    addresses: List[Address],
+    state    : DataState[NFTUpdatesState, NFTUpdatesCalculatedState]
+  ): F[DataApplicationValidationErrorOr[Unit]] = {
+    val validateAddress = Async[F].delay(validateProvidedAddress(addresses, update.fromAddress))
```
sdavidp: Same comment as above

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/validations/Validations.scala
```diff
@@ -1,98 +1,128 @@
...
   }
 
-  def transferNFTValidationsWithSignature(update: TransferNFT, addresses: List[Address], state: DataState[NFTUpdatesState, NFTUpdatesCalculatedState]): IO[DataApplicationValidationErrorOr[Unit]] = {
-    val validateAddress = validateProvidedAddress(addresses, update.fromAddress).pure[IO]
+  def transferNFTValidationsWithSignature[F[_] : Async](
+    update   : TransferNFT,
+    addresses: List[Address],
+    state    : DataState[NFTUpdatesState, NFTUpdatesCalculatedState]
+  ): F[DataApplicationValidationErrorOr[Unit]] = {
+    val validateAddress = Async[F].delay(validateProvidedAddress(addresses, update.fromAddress))
```
sdavidp: and again

sdavidp (commented):
Reviewed the changes since last review. Mostly just suggested tweaks but there's a few questions as well.

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/validations/Validations.scala
sdavidp: In case the difference is not obvious, here's a sample test. If you uncomment the 2nd assertion you'll see it fail and see the difference.
```
test("productR") { _ =>
  import cats.syntax.all._
  import org.tessellation.currency.dataApplication.{DataApplicationValidationError, Noop}
  import org.tessellation.currency.dataApplication.dataApplication.DataApplicationValidationErrorOr

  case object Hoop extends DataApplicationValidationError {
    val message = "hoop"
  }

  val hoop : DataApplicationValidationErrorOr[Unit] = Hoop.invalidNec[Unit]
  val hoopF = hoop.pure[IO]

  val noop: DataApplicationValidationErrorOr[Unit] = Noop.invalidNec[Unit]
  val noopF = noop.pure[IO]

  for {
    prodF: DataApplicationValidationErrorOr[Unit] <- hoopF.productR(noopF)
    prodV: DataApplicationValidationErrorOr[Unit] = hoop.productR(noop)
    prodM: DataApplicationValidationErrorOr[Unit] <- hoopF.map2(noopF)(_.productR(_))
  } yield expect.all(
    prodM == prodV
//    , prodF == prodV
  )
}
```

marcus-girolneto: Thanks, updated

examples/DataApi-NFT/template/modules/data_l1/src/main/scala/com/my/nft_example/data_l1/Main.scala
marcus-girolneto: Fixed

examples/DataApi-NFT/template/modules/l0/src/main/scala/com/my/nft_example/l0/Main.scala
marcus-girolneto: Fixed

examples/DataApi-NFT/template/modules/l0/src/main/scala/com/my/nft_example/l0/custom_routes/CustomRoutes.scala
marcus-girolneto: Updated

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/Utils.scala
marcus-girolneto: This is good

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/Utils.scala
marcus-girolneto: Got it

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/combiners/Combiners.scala
marcus-girolneto: Got it

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/deserializers/Deserializers.scala
marcus-girolneto: This is the `jawn.decode`?

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/validations/Validations.scala
marcus-girolneto: Removed the delay here

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/validations/Validations.scala
marcus-girolneto: Removed

examples/DataApi-NFT/template/modules/shared_data/src/main/scala/com/my/nft_example/shared_data/LifecycleSharedFunctions.scala
```diff
@@ -22,13 +22,13 @@ object LifecycleSharedFunctions {
   ): F[DataApplicationValidationErrorOr[Unit]] =
     update match {
       case mintCollection: MintCollection =>
-        mintCollectionValidations(mintCollection, None)
+        Async[F].delay(mintCollectionValidations(mintCollection, None))
```
sdavidp: Just one `.delay` containing the match expression?

sdavidp (approved):
LGTM. There's still a couple places where `.delay` is repeated for all cases of a `match` when you can just put the whole expression in one `.delay`

PR #14, opened 2024-04-11
Custom transaction validator example

kpudlik:


marcus-girolneto (approved):
LGTM. However, I think we should wait for the merge, I mean we should merge only when the official tessellation version release

marcinwadon (approved):

kpudlik:
> LGTM. However, I think we should wait for the merge, I mean we should merge only when the official tessellation version release

Yeah, I agree. For now, I hardcoded 2.6.0 version as it will probably be the next one but if the change is deployed in a different version then we should adjust this PR.

examples/custom-transaction-validation/modules/l1/src/main/scala/com/my/custom_validator/l1/Main.scala
```diff
@@ -0,0 +1,32 @@
...
+    "custom-transaction-validation-l1",
+    "custom-transaction-validation L1 node",
+    ClusterId(UUID.fromString("517c3a05-9219-471b-a54c-21b7d72f4ae5")),
+    tessellationVersion = TessellationVersion.unsafeFrom(BuildInfo.version),
+    metagraphVersion = MetagraphVersion.unsafeFrom(BuildInfo.version)
+  ) {
+  override def transactionValidator: Option[CustomContextualTransactionValidator] = Some {
+    (hashedTransaction: Hashed[Transaction], context: TransactionValidatorContext) =>
+      Either.cond(
+        context.currentOrdinal.value < 10L || hashedTransaction.fee.value % 2 == 0,
```
AlexBrandes: I think we should make this a more realistic example. I can't think of a real use case for a fee that must be an even number. How about we replace the even number requirement with a minimum fee?

examples/custom-transaction-validation/README.md
```diff
@@ -0,0 +1,15 @@
+# Metagraph - Custom transaction validation
+
+This example validates incoming transactions with the custom validator.
```
AlexBrandes:
```suggestion
This example provides a sample implementation of a currency metagraph with a custom transaction validator. In this case, the validator prevents transactions below a minimum fee from being accepted.

The CurrencyL1App provides `transactionValidator` as an overridable method to introduce custom validation logic to token transactions. Throwing an error from this method will reject the transaction and prevent its execution. 
```

examples/custom-transaction-validation/README.md
```diff
@@ -0,0 +1,15 @@
...
+The validation criteria are as following:
+- If snapshot ordinal < 10 then pass
+- If snapshot ordinal >= 10 then pass only if fee is an even number
+
```
AlexBrandes:
```suggestion

The following validation criteria are implemented in this example:
- If snapshot ordinal < 10 then pass
- If snapshot ordinal >= 10 then pass only if fee is greater than 10
```

examples/custom-transaction-validation/README.md
```diff
@@ -0,0 +1,15 @@
...
+## Template
+This project contains dependencies of the Tessellation on version `v2.6.0`
+
+At this example you can take a look at the file:
```
AlexBrandes:
```suggestion
The primary implementation of this example is found in: 
```

AlexBrandes (changes requested):

examples/custom-transaction-validation/modules/l1/src/main/scala/com/my/custom_validator/l1/Main.scala
kpudlik: Sure, I will replace it then :+1:

kpudlik:
@AlexBrandes @IPadawans Marcin has released 2.6.0 including that change so I think it is mergable.

AlexBrandes (approved):
LGTM

PR #22, opened 2024-07-25
Adding social metagraph example

marcus-girolneto:
### Changes
+ Adding new metagraph example: Social
+ This metagraph should work like a twitter clone (partially). In this metagraph your user is defined by your signing wallet and you can: CreatePost, EditPost, DeletePost, Subscribe
+ This metagraph offers support to external storage mechanisms, most specifically Postgres. This means the calculated state will be persisted in Postgres
+ The README was updated with instructions to run using Euclid

### Tests
+ It worked as expected:
Captura de Tela 2024-07-25 às 16 55 04


examples/social/README.md
```diff
@@ -0,0 +1,173 @@
...
+-   `serializeUpdate`
+-   `deserializeUpdate`
+-   `setCalculatedState`
+-   `getCalculatedState`
+-   `hashCalculatedState`
+-   `routes`
+
+For a more detailed understanding, please refer to the [complete documentation](https://docs.constellationnetwork.io/sdk/frameworks/currency/data-api) on the Data API.
+
+### Lifecycle Functions
```
AlexBrandes: Let's not repeat these lifecycle functions on each example. I think we can remove here.

examples/social/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+#### -> `getCalculatedState`
+
+-   This function retrieves the calculated state.
+
+#### -> `hashCalculatedState`
+
+-   This function creates a hash of the calculatedState to be validated when rebuilding this state, in case of restarting the metagraph.
+
+#### -> `routes`
```
AlexBrandes: This part still makes sense/should stay.

examples/social/README.md
```diff
@@ -0,0 +1,173 @@
+Metagraph - Data API - Social
```
AlexBrandes: Data API is now called Data Application

examples/social/README.md
```diff
@@ -0,0 +1,173 @@
...
+
+This example demonstrates a basic social media use case using the Data API. In the example, a client can send four types of signed data updates to a metagraph: one to create a post, another to edit a post, another to delete a post, and another to subscribe to a user. These updates are validated before being merged into the snapshot state.
+
+Here's how the social media system works:
+
+-   Each user can create, edit, and delete their own posts.
+-   Each user can subscribe to other users to follow their posts.
+-   Posts and subscriptions are validated to ensure their integrity and consistency.
+-   Users cannot create duplicate posts with the same content.
+
```
AlexBrandes: One of the main goals of this example is to show an external data store being used to store calculated state. We should describe in detail how that's implemented here.

examples/social/README.md
```diff
@@ -0,0 +1,173 @@
...
+#### -> `hashCalculatedState`
+
+-   This function creates a hash of the calculatedState to be validated when rebuilding this state, in case of restarting the metagraph.
+
+#### -> `routes`
+
+Customizes routes for our application.
+
+In this example, the following endpoints are implemented:
+
```
AlexBrandes, line 57: Let's add some basic resource endpoints to make this easier to explore:
 - /users: endpoint that lists users that have posts. 
 - /posts: endpoint that returns all posts

examples/social/README.md
```diff
@@ -0,0 +1,173 @@
...
+-   GET `<metagraph l0 url>/data-application/users/:user_id/posts`: Returns all user posts.
+-   GET `<metagraph l0 url>/data-application/posts/:user_id/subscriptions`: Returns the subscriptions of a user.
```
AlexBrandes: I don't think `posts` belongs in the url here. Just `/users/:user_id/subscriptions` is enough. 

```suggestion
-   GET `<metagraph l0 url>/data-application/users/:user_id/subscriptions`: Returns the subscriptions of a user.
```

examples/social/README.md
```diff
@@ -0,0 +1,173 @@
...
+-   GET `<metagraph l0 url>/data-application/users/:user_id/feed`: Returns the user feed.
```
AlexBrandes:
```suggestion
-   GET `<metagraph l0 url>/data-application/users/:user_id/feed`: Returns the user's feed of posts from their subscriptions.
```

examples/social/modules/shared_data/src/main/scala/com/my/metagraph_social/shared_data/calculated_state/CalculatedStateService.scala
```diff
@@ -0,0 +1,61 @@
...
+  ): F[CalculatedStateService[F]] =
+    Ref.of[F, CalculatedState](CalculatedState.empty).map { stateRef =>
+      new CalculatedStateService[F] {
+        override def get: F[CalculatedState] = stateRef.get
+
+        override def set(
+          snapshotOrdinal: SnapshotOrdinal,
+          state          : SocialCalculatedState
+        ): F[Boolean] = for {
+          _ <- externalStorageService.set(snapshotOrdinal, state)
```
scasplte2: I do not believe this is an atomic operation at the moment, so the external store may fail to update while the internal memory is updated, thus getting out of sync. I would need to double-check but I think you want to use `stateRef.modify` and to have the storage update side-effect as part of that modification to ensure that the operation is atomic.

scasplte2 (changes requested):

examples/social/modules/shared_data/src/main/scala/com/my/metagraph_social/shared_data/calculated_state/CalculatedStateService.scala
scasplte2: I think the atomic version would be
```scala
 override def set(
          snapshotOrdinal: SnapshotOrdinal,
          state          : SocialCalculatedState
        ): F[Boolean] = stateRef.modify { currentState =>
        val effect = externalStorageService.set(snapshotOrdinal, state)
        val currentCalculatedState = currentState.state
            val updatedUsers = state.users.foldLeft(currentCalculatedState.users) {
              case (acc, (address, value)) =>
                acc.updated(address, value)
            }
         val state = CalculatedState(snapshotOrdinal, SocialCalculatedState(updatedUsers))
         
         (state, effect)
      }.as(true)
        .handleErrorWith { err =>
          Logger[F].warn(s"CalculatedState failed to update with error: $err") *> false.pure[F]
        }
```

marcus-girolneto: Updated

examples/social/README.md
marcus-girolneto: Updated

examples/social/README.md
marcus-girolneto: Updated

examples/social/README.md
marcus-girolneto: Updated

examples/social/README.md
marcus-girolneto: Updated

examples/social/README.md
marcus-girolneto: Updated

examples/social/README.md
marcus-girolneto: Updated

examples/social/README.md
marcus-girolneto: Updated

scasplte2 (approved):
LGTM :rocket:

PR #23, opened 2024-07-25
Updated README - added example descriptions and functionality

AlexBrandes:


scasplte2:
@AlexBrandes This looks good as is for the current examples but would you like to update this with the social or todo examples? I'm not sure what the differentiation could be on the todo one other than being tied to a tutorial though.

AlexBrandes:
> @AlexBrandes This looks good as is for the current examples but would you like to update this with the social or todo examples? I'm not sure what the differentiation could be on the todo one other than being tied to a tutorial though.

@scasplte2 Once we have a final video link for the todo tutorial then I think we should definitely include it in the README. Can it be merged into main or does it need to stay on a branch to preserve the commit structure? Maybe some combination of both so that it's discoverable but we also prevent the commits from getting buried.

PR #24, opened 2024-07-25
Todo example metagraph

scasplte2:
## Purpose
Adds a todo metagraph example which is a simple task assignment tool used as a build tutorial for developing metagraphs on the Constellation Network.

## Testing
Ran local metagraph and verified expected behavior with generated transactions and postman collection


examples/todo/modules/data_l1/src/main/scala/com/my/data_l1/DataL1Service.scala
```diff
@@ -0,0 +1,113 @@
...
+      dataApplicationL1Service = makeBaseApplicationL1Service(validator)
+    } yield dataApplicationL1Service
+
+  private def makeBaseApplicationL1Service[F[+_]: Async: JsonSerializer](
+    validator: DataL1Validator[F, TodoUpdate, OnChain]
+  ): BaseDataApplicationL1Service[F] =
+    BaseDataApplicationL1Service[F, TodoUpdate, OnChain, CalculatedState](
+      new DataApplicationL1Service[F, TodoUpdate, OnChain, CalculatedState] {
+
+        override def serializeState(state: OnChain): F[Array[Byte]] =
```
marcinwadon, line 42: do you override something or you provide an implementation? I think you override nothing so the `override` modifier is unnecessary.

examples/todo/modules/data_l1/src/main/scala/com/my/data_l1/DataL1Service.scala
```diff
@@ -0,0 +1,113 @@
...
+            _.fold(
+              err => err.invalidNec[Unit].pure[F],
+              onchain => validator.verify(onchain, update)
+            )
+          }
+
+        override def combine(
+          state:   DataState[OnChain, CalculatedState],
+          updates: List[Signed[TodoUpdate]]
+        )(implicit context: L1NodeContext[F]): F[DataState[OnChain, CalculatedState]] = state.pure[F]
```
marcinwadon, line 107: how does it work if you don't apply updates onto the previous state?

examples/todo/modules/l0/src/main/scala/com/my/metagraph_l0/Main.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  override def dataApplication: Option[Resource[IO, BaseDataApplicationL0Service[IO]]] = (for {
+    config                                           <- ApplicationConfigOps.readDefault[IO].asResource
+    implicit0(logger: SelfAwareStructuredLogger[IO]) <- Slf4jLogger.create[IO].asResource
+    implicit0(supervisor: Supervisor[IO])            <- Supervisor[IO]
+    implicit0(json2bin: JsonSerializer[IO])          <- JsonBinaryCodec.forSync[IO].asResource
+    implicit0(hasher: Hasher[IO])                    <- Hasher.forJson[IO].pure[IO].asResource
+    implicit0(sp: SecurityProvider[IO])              <- SecurityProvider.forAsync[IO]
+//    _                                                <- makePeriodicDaemon.asResource
+//    _                                                <- makePeriodicConfigDaemon(config).asResource
```
marcinwadon: ?

examples/todo/modules/shared-data/src/main/resources/logback.xml
```diff
@@ -0,0 +1,19 @@
...
+            <pattern>%d{HH:mm:ss.SSS} [%thread] %highlight(%-5level) %cyan(%logger{15}) - %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <root level="INFO">
+        <appender-ref ref="STDOUT" />
+    </root>
+
+    <logger name="org.tessellation" level="DEBUG" />
+    <logger name="xyz.kd5ujc" level="DEBUG" />
```
marcinwadon: ?

examples/todo/modules/shared-data/src/main/scala/com/my/shared_data/lib/OrderedOps.scala
```diff
@@ -0,0 +1,24 @@
...
+          def order(todo: TodoUpdate): Int = todo match {
+            case _: CreateTask   => 0
+            case _: ModifyTask   => 1
+            case _: CompleteTask => 2
+            case _: RemoveTask   => 3
+          }
+
+          order(x) compareTo order(y)
+        }
+      }
```
marcinwadon, line 22: So, two todos are equally the same if they are of the same type?

examples/todo/modules/shared-data/src/main/scala/com/my/shared_data/lib/package.scala
```diff
@@ -0,0 +1,65 @@
...
+    implicit class CurrencyIncrementalSnapshotOps[F[_]: Sync: JsonSerializer](cis: CurrencyIncrementalSnapshot) {
+
+      def countUpdates: F[Long] = getBlocks.map(_.map(_.updates.size.toLong).sum)
+
+      def getUpdates[U <: DataUpdate: ClassTag]: F[List[Signed[U]]] =
+        getBlocks
+          .map {
+            _.flatMap { block =>
+              block.updates.toList.collect { case s: U =>
+                val flightUpdate = s.value.asInstanceOf[U]
```
marcinwadon: leftover?

examples/todo/project/Dependencies.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  def decline(artifact: String = ""): ModuleID =
+    "com.monovore" %% {
+      if (artifact.isEmpty) "decline" else s"decline-$artifact"
+    } % V.decline
+
+  object Libraries {
+    val tessellationNodeShared = tessellation("node-shared")
+    val tessellationCurrencyL0 = tessellation("currency-l0")
+    val tessellationCurrencyL1 = tessellation("currency-l1")
```
marcinwadon, line 24: there should be a single dependency to tessellation sdk

examples/todo/project/plugins.sbt
```diff
@@ -0,0 +1,20 @@
...
+addSbtPlugin("com.codecommit" % "sbt-github-packages" % "0.5.3")
+addSbtPlugin("com.timushev.sbt" % "sbt-updates" % "0.6.3")
+addSbtPlugin("com.thesamet" % "sbt-protoc" % "1.0.6") //https://github.com/thesamet/sbt-protoc
+
+// https://index.scala-lang.org/scalapb/scalapb/artifacts/compilerplugin
+// NOTE! This must be here and not in build.sbt otherwise the `scalapb.gen()` isn't seen
+libraryDependencies ++= Seq(
+  "com.thesamet.scalapb" %% "compilerplugin"           % "0.11.13",
+  "com.thesamet.scalapb" %% "scalapb-validate-codegen" % "0.3.4"
+)
```
marcinwadon: do we need the protobuf for that example?

examples/todo/modules/data_l1/src/main/scala/com/my/data_l1/DataL1Service.scala
scasplte2: The DataL1 never invokes the combine function but the inheritance from Tessellation is such that an implementation must be provided so this is just the simplest implementation.

examples/todo/modules/l0/src/main/scala/com/my/metagraph_l0/Main.scala
scasplte2: You're right, this could be removed. I added them as a part of the tutorial for the hackathon webinar.

examples/todo/modules/shared-data/src/main/resources/logback.xml
scasplte2: thank you, a relic from having ported this from a personal metagraph project

examples/todo/modules/shared-data/src/main/scala/com/my/shared_data/lib/OrderedOps.scala
scasplte2: I was just aiming for grouping all like Tasks together when applying them onto the DataState. I was not concerned with ordering between like Tasks

examples/todo/modules/shared-data/src/main/scala/com/my/shared_data/lib/package.scala
scasplte2: oof.. yes, thank you

examples/todo/project/plugins.sbt
scasplte2: no, I thought I had removed all the `scalapb` deps. Thank you for pointing out.

examples/todo/modules/data_l1/src/main/scala/com/my/data_l1/DataL1Service.scala
scasplte2: I'm not sure I understand. This does override the `serializeState` method from `BaseDataApplicationService` which the `BaseDataApplicationL1Service` [inherits from](https://github.com/Constellation-Labs/tessellation/blob/36e8a0fa551842000823584fad9c951474c065dc/modules/shared/src/main/scala/org/tessellation/currency/dataApplication/package.scala#L110).

examples/todo/project/Dependencies.scala
scasplte2: I can replace `node-shared` with `sdk` but am unable to use just `tessellation-sdk` as it does not include the dependencies for `currency-l1` and `currency-l0`

marcinwadon: Hmm, then we should add both to the `sdk`.
The `sdk` module should be the only one that external projects would depend on.

examples/todo/modules/shared-data/src/main/scala/com/my/shared_data/lib/OrderedOps.scala
marcinwadon: Ok, but it's a default implicit so it may arise additional errors with sorting/ordering (theoretically, didn't check). Take a look how cats derive `Eq` from `Order`.

We encountered many times in the past that for the ordering like that:
```
case Snapshot(ordinal: Long, a: Foo, b: Bar, c: Baz)

object Snapshot {
  implicit val order: Order[Snapshot] = ... a.ordinal > b.ordinal // (or sth like that, based on one field only)
}
```

snapshots with the same ordinal would be considered equal, even though `a`, `b` and `c` properties are different.

examples/todo/modules/data_l1/src/main/scala/com/my/data_l1/DataL1Service.scala
marcinwadon: as far as I see, `BaseDataApplicationService` does not implement any of these methods.
image

so you don't override the default implementation, but you provide the first implementation there. hence, you "override" nothing.

marcinwadon: Oh, I see. You have implemented the `data l1` here only. I think you should provide a single `DataApplicationService` and reuse it for each layer 🤔 you should not implement each layer independently 🤔

examples/todo/modules/data_l1/src/main/scala/com/my/data_l1/DataL1Service.scala
marcinwadon: yeah, according to the previous comment, you shouldn't make separate implementations for each layer, but you should provide a single implementation and reuse it for each layer. so that, you don't need any dummy implementations

examples/todo/modules/data_l1/src/main/scala/com/my/data_l1/DataL1Service.scala
scasplte2: oh, I see, I believe I misunderstood your comment at first. You are correct that there are no other implementations that are being overridden so it is not strictly needed in this case. I was under the impression that it was common practice to use the override modifier to signify that an implementation is required due to inheritance.

examples/todo/modules/data_l1/src/main/scala/com/my/data_l1/DataL1Service.scala
scasplte2: I don't understand how would be possible to provide a single implementation for reuse between the Data-L1 and L0 services? Tessellation makes the distinction of `BaseDataApplicationL0Service` and `BaseDataApplicationL1Service` where the `Context` type constraint from `BaseDataApplicationService` is concretely tied to `L0NodeContext` and `L1NodeContext`. I agree that functions within each service could share functional definitions by pointing to some object as done in [Dor](https://github.com/Constellation-Labs/dor-metagraph/blob/d85e808d12bbb491fce8e789fe72f2d736cc4328/metagraph/modules/data_l1/src/main/scala/com/my/dor_metagraph/data_l1/Main.scala#L49). But in the specific case of the `combine(...)(implicit L1NodeContext)` this function is not called within Tessellation but is required by inheritance.

It would be my preference to not include these extraneous methods within the definition of each service as well but from what I can see in Tessellation, that seems like a significant refactor of the Data Application inheritance structure and overall usage.

PR #26, opened 2024-07-30
Sample metagraph using estimate-fee and validateFee

sdavidp:


examples/.DS_Store
marcinwadon, line 1: I think you've added this before .gitignore

examples/validate-fee/project/Dependencies.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  def decline(artifact: String = ""): ModuleID =
+    "com.monovore" %% {
+      if (artifact.isEmpty) "decline" else s"decline-$artifact"
+    } % V.decline
+
+  object Libraries {
+    val tessellationNodeShared = tessellation("node-shared")
+    val tessellationCurrencyL0 = tessellation("currency-l0")
+    val tessellationCurrencyL1 = tessellation("currency-l1")
```
marcinwadon, line 23: if possible, the only dependency to tessellation should be the `sdk`

examples/validate-fee/sample-ui/src/app/polls/[poll_id]/actions.ts
```diff
@@ -0,0 +1,59 @@
...
+          status: response.status,
+          content: await response.text()
+        }
+      }
+    };
+  }
+
+  const responseData: [string, string] = await response.json();
+
+  console.log(responseData);
```
Comment (marcinwadon) [inline examples/validate-fee/sample-ui/src/app/polls/[poll_id]/actions.ts]: leftover or intented?

examples/validate-fee/sample-ui/src/app/polls/[poll_id]/form.tsx
```diff
@@ -0,0 +1,197 @@
...
+  const buildAndSignPayload = async () => {
+    const validatedValues = FormVoteSchema.safeParse(getValues());
+
+    if (!validatedValues.data) {
+      throw new Error('Inconsistency Error');
+    }
+
+    const values = validatedValues.data;
+
+    console.log(values);
```
Comment (marcinwadon) [inline examples/validate-fee/sample-ui/src/app/polls/[poll_id]/form.tsx]: leftover?

examples/validate-fee/sample-ui/src/app/polls/create/actions.ts
```diff
@@ -0,0 +1,63 @@
...
+import { PollSchema } from '../../../schemas';
+import { MetagraphBaseURLs } from '../../../consts';
+
+const CreatePollSchema = PollSchema.extend({
+  signedPayload: z
+    .string({ invalid_type_error: 'Invalid signed payload' })
+    .regex(
+      /^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/,
+      'Invalid base64 character'
+    )
```
marcinwadon: I see this regex third or fourth time. Maybe it's worth to put it in one place and reuse?

examples/validate-fee/sample-ui/src/app/polls/create/form.tsx
```diff
@@ -0,0 +1,203 @@
...
+import { z } from 'zod';
+import { useEffect } from 'react';
+import { toast } from 'react-toastify';
+
+import { ButtonLink } from '../../../components/Button/ButtonLink/component.tsx';
+import { useWalletProvider } from '../../../providers/index.ts';
+import { PollSchema } from '../../../schemas/index.ts';
+import { buildRegisterField } from '../../../utils/forms.ts';
+import { Button, Card, Input } from '../../../components/index.ts';
+import { useToastAction } from '../../../utils/use_toast_action.ts';
```
marcinwadon: it's better to use aliases like `@schemas/`, `@components/` or `@utils/` than `../../../`

examples/validate-fee/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Template
+
+Primary code for the example can be found in the following files:
+
+`modules/data_l1/src/main/scala/com/my/validate_fee/Main.scala`
+
+## Scripts
+
+`scripts/send_data_transaction.js` is a simple script to send an initial message to the metagraph.
```
AlexBrandes: I don't see a scripts directory here. Does that need to get added to git?

examples/validate-fee/sample-ui/.eslintignore
```diff
@@ -0,0 +1,43 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
AlexBrandes: Let's remove the sample-ui if it's not actually used for the example.

examples/validate-fee/modules/shared_data/src/main/scala/com/my/validate_fee/shared_data/types/Types.scala
```diff
@@ -0,0 +1,28 @@
...
+import org.tessellation.currency.schema.feeTransaction.FeeTransaction
+import org.tessellation.schema.balance.Amount
+import eu.timepit.refined.auto._
+
+object Types {
+  @derive(decoder, encoder)
+  sealed trait SampleUpdate extends DataUpdate
+
+  val updateTypeOneFee = Amount(123L)
+  val updateTypeTwoFee = Amount(1000L)
```
AlexBrandes: This example is good but maybe we could extend the functionality a bit. Can we make the type two fee variable in some way?

Like, maybe have the fee make use of calculated state to determine the amount based on usage? So the 1st request from a wallet would be a fee of 1, 2nd fee of 2, etc. 

Or else we could use snapshot ordinal floor 10, or something like that to make it variable: `fee = floor(ordinal/10)`

Open to ideas but it would be great if we could show a slightly more advanced fee functionality (like one of the above) alongside a basic static fee.

examples/validate-fee/modules/shared_data/src/main/scala/com/my/validate_fee/shared_data/errors/Errors.scala
```diff
@@ -0,0 +1,25 @@
...
+
+  implicit class DataApplicationValidationTypeOps[E <: DataApplicationValidationError](err: E) {
+    def invalid: DataApplicationValidationType = err.invalidNec[Unit]
+
+    def unlessA(cond: Boolean): DataApplicationValidationType = if (cond) valid else invalid
+
+    def whenA(cond: Boolean): DataApplicationValidationType = if (cond) invalid else valid
+  }
+
+  case class UnexpectedAmount(name: String, expected: Amount, actual: Amount) extends DataApplicationValidationError {
```
AlexBrandes: I think this should use the `DataApplicationFeeError` in place of `DataApplicationValidationError`

examples/validate-fee/modules/shared_data/src/main/scala/com/my/validate_fee/shared_data/Validators.scala
```diff
@@ -0,0 +1,21 @@
...
+import org.tessellation.currency.dataApplication.dataApplication.DataApplicationValidationErrorOr
+import org.tessellation.schema.SnapshotOrdinal
+
+object Validators {
+  def validateSampleUpdateFee[F[_]: Applicative](unused: SnapshotOrdinal, update: SampleUpdate): F[DataApplicationValidationErrorOr[Unit]] =
+    update
+      .pure[F]
+      .map {
+        case UpdateTypeOne(name, fee) =>
+          UnexpectedAmount(name, updateTypeOneFee, fee.amount).unlessA(updateTypeOneFee === fee.amount)
```
AlexBrandes: I think less than or equal to would make a better comparison here.

examples/validate-fee/README.md
sdavidp: For some reason,  the `.gitignore` (copied from the voting-poll example) contains `scripts/`.
@IPadawans Is that by design?

marcus-girolneto: You can remove it from gitignore, it should contain the scripts folder. I've already removed it from the new project.

scasplte2: the todo example put a .gitignore at the root of the project (since that is where the .git resides) but I'm realizing now that we may want the individual gitignore within each project so euclid to copy?

examples/validate-fee/modules/shared_data/src/main/scala/com/my/validate_fee/shared_data/Validators.scala
sdavidp: @AlexBrandes Can you clarify? Is it ok or an error if the given fee is less than expected?

examples/validate-fee/modules/shared_data/src/main/scala/com/my/validate_fee/shared_data/errors/Errors.scala
sdavidp: Done; will appear in the next push.

examples/validate-fee/modules/shared_data/src/main/scala/com/my/validate_fee/shared_data/types/Types.scala
sdavidp: > This example is good but maybe we could extend the functionality a bit. Can we make the type two fee variable in some way?
> 
> Like, maybe have the fee make use of calculated state to determine the amount based on usage? So the 1st request from a wallet would be a fee of 1, 2nd fee of 2, etc.

That was what I initially had but thought that would be criticized as not necessary.

> Or else we could use snapshot ordinal floor 10, or something like that to make it variable: `fee = floor(ordinal/10)`

One of my variations used the ordinal to generate the fee but I ran into issues with the ordinal not being the same when `estimate-fee` is called vs validation later on in the flow so the update failed validation. I could be mistaken on the problem though. I'll look into it.

sdavidp: @AlexBrandes Is calculated state persisted between node restarts? (This is metagraph-dependent, isn't it?) If not, won't that make fee calculations based on calculated state non-deterministic?

examples/validate-fee/sample-ui/src/app/polls/create/form.tsx
sdavidp: Deleted sample-ui

examples/validate-fee/sample-ui/src/app/polls/create/actions.ts
sdavidp: Deleted sample-ui

examples/validate-fee/sample-ui/src/app/polls/[poll_id]/form.tsx
Comment (sdavidp) [reply examples/validate-fee/sample-ui/src/app/polls/[poll_id]/form.tsx]: Deleted sample-ui

examples/.DS_Store
sdavidp: Removed.

examples/validate-fee/sample-ui/.eslintignore
sdavidp: Deleted sample-ui

examples/validate-fee/project/Dependencies.scala
sdavidp: @marcinwadon I'll look into this more later but replacing these with "sdk" causes the euclid build to fail. It's not able to download/see the sdk jar, which I do see in my `.ivy/` directory.

marcinwadon: Sure. I think it may be the euclid that does not assembly the sdk module but only those you used.

examples/validate-fee/modules/shared_data/src/main/scala/com/my/validate_fee/shared_data/calculated_state/DataCalculatedStateService.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  def make[F[_] : Async, A <: DataCalculatedState : Monoid: Encoder]: F[DataCalculatedStateService[F, A]] = {
+    Ref.of[F, (SnapshotOrdinal, A)](SnapshotOrdinal.MinValue -> Monoid[A].empty).map { stateRef =>
+      new DataCalculatedStateService[F, A] {
+        override def getCalculatedState: F[(SnapshotOrdinal, A)] = stateRef.get
+
+        override def setCalculatedState(snapshotOrdinal: SnapshotOrdinal, state: A): F[Boolean] =
+          stateRef.modify { case (ordinal, currentState) =>
+            ((snapshotOrdinal, currentState |+| state), true)
+          }
```
sdavidp: What is the relationship between new ordinal and that in the map?
Are there conditions when we should or should not update?
E.g. should we only update if snapshotOrdinal > ordinal ?
Should we keep all updates when ordinals differ?

examples/validate-fee/modules/shared_data/src/main/scala/com/my/validate_fee/shared_data/Validators.scala
AlexBrandes: It's an error if the given fee is less than expected. It shouldn't be an error if the given fee is more than expected. If the cost to do some action is 10 and I pay 20, it's still enough and should be accepted.

examples/validate-fee/modules/shared_data/src/main/scala/com/my/validate_fee/shared_data/types/Types.scala
AlexBrandes: Calculated state is recalculated after each restart so it's available in the same form/persisted.

examples/validate-fee/modules/shared_data/src/main/scala/com/my/validate_fee/shared_data/calculated_state/DataCalculatedStateService.scala
```diff
@@ -0,0 +1,42 @@
...
+object DataCalculatedStateService {
+
+  def make[F[_] : Async, A <: DataCalculatedState : Monoid: Encoder]: F[DataCalculatedStateService[F, A]] = {
+    Ref.of[F, (SnapshotOrdinal, A)](SnapshotOrdinal.MinValue -> Monoid[A].empty).map { stateRef =>
+      new DataCalculatedStateService[F, A] {
+        override def getCalculatedState: F[(SnapshotOrdinal, A)] = stateRef.get
+
+        override def setCalculatedState(snapshotOrdinal: SnapshotOrdinal, state: A): F[Boolean] =
+          stateRef.modify { case (_, currentState) =>
+            ((snapshotOrdinal, currentState |+| state), true)
```
scasplte2: I'm not sure if combining the current state and previous state here is needed. It seems like `DataApplicationSnapshotAcceptanceManager.consumeSignedMajorityArtifact` is the location where this set function is modified (other than download operations where this is just set). Based on this portion of the function, then it seems like if the `CalculatedState` is modified correctly in the Combiner functions then shouldn't state already be a super set (or contain the data) that is within current state?

```scala
da => OptionT
  .liftF(da.blocks.traverse(service.deserializeBlock).map(_.flatMap(_.toOption)))
  .flatMapF { dataBlocks =>
    artifact.ordinal.partialPrevious.flatTraverse(lastOrdinal =>
      accept(maybeLastDataApplication, dataBlocks, lastOrdinal, artifact.ordinal)
    )
  }
  .map(_.calculatedState)
  .semiflatMap(expectCalculatedStateHash(da.calculatedStateProof))
  .semiflatTap(service.setCalculatedState(artifact.ordinal, _))
  .semiflatTap(calculatedStateStorage.write(artifact.ordinal, _)(service.serializeCalculatedState))
```

I believe this function can be simplified to 
```scala
override def set(checkpoint: Checkpoint[S]): F[Boolean] =
  state
    .set(checkpoint)
    .as(true)
    .handleErrorWith { err =>
      Logger[F].warn(s"Checkpoint set failed with error: $err") *> false.pure[F]
    }
```

examples/validate-fee/project/Dependencies.scala
scasplte2: That is correct. Currently Euclid assembles and publishes[ these packages](https://github.com/Constellation-Labs/euclid-development-environment/blob/bc8f08e4b7b90161171b325e1e524c7c6f86c146/infra/docker/metagraph-ubuntu/Dockerfile#L35)
```
 sbt kernel/publishM2 shared/publishM2 keytool/publishM2 nodeShared/publishM2 dagL1/publishM2 currencyL0/publishM2 currencyL1/publishM2;
 
 sbt wallet/assembly dagL0/assembly dagL1/assembly keytool/assembly;
 ```

Comment (scasplte2) [reply examples/validate-fee/project/Dependencies.scala:23]: Looking at recent releases on Tessellation, I'm not sure if the [PR](https://github.com/Constellation-Labs/tessellation/pull/912) modifying  the `sdk` modules has been merged into a release yet.

File: examples/validate-fee/modules/shared_data/src/main/scala/com/my/validate_fee/shared_data/calculated_state/DataCalculatedStateService.scala (hunk 12 above)
Comment (sdavidp) [reply examples/validate-fee/modules/shared_data/src/main/scala/com/my/validate_fee/shared_data/calculated_state/DataCalculatedStateService.scala]: @scasplte2 If understand correctly, are you suggesting `setCalculatedState` simply overwrite the existing state with the one given?

Comment (scasplte2) [reply examples/validate-fee/modules/shared_data/src/main/scala/com/my/validate_fee/shared_data/calculated_state/DataCalculatedStateService.scala]: yes, the combiner takes care of generating the new state that is associated to the ordinal so I believe it is redundant to do it again here.

File: examples/validate-fee/modules/l0/src/main/resources/logback.xml [hunk 13]
```diff
@@ -0,0 +1,13 @@
...
+        <encoder>
+            <pattern>%d{HH:mm:ss.SSS} [%thread] %highlight(%-5level) %cyan(%logger{15}) - %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <root level="INFO">
+        <appender-ref ref="STDOUT" />
+    </root>
+
+    <logger name="com.my.validate_fee" level="INFO" />
```
scasplte2, line 12: Could you add the exceptions to block http4s messages here as in Tessellation? I spun up this example and was able to send the two transactions (seemingly successfully) but I can't really tell because the app logs are inundated with the messages from `ResponseLogger, RequestLogger, etc.`

examples/validate-fee/modules/data_l1/src/main/resources/logback.xml
```diff
@@ -0,0 +1,13 @@
...
+        <encoder>
+            <pattern>%d{HH:mm:ss.SSS} [%thread] %highlight(%-5level) %cyan(%logger{15}) - %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <root level="INFO">
+        <appender-ref ref="STDOUT" />
+    </root>
+
+    <logger name="com.my.validate_fee" level="INFO" />
```
scasplte2, line 12: Same request here about replicating the Tessellation log exclusions

scasplte2 (changes requested):
Application spins up and both transactions were received by the metagraph. However, I am not sure where to look to ensure successful operation as the logs themselves are quite busy.

examples/validate-fee/modules/l0/src/main/resources/logback.xml
sdavidp: done

examples/validate-fee/modules/data_l1/src/main/resources/logback.xml
sdavidp: done

scasplte2 (approved):

PR #33, opened 2024-09-30
feat: adding nft-fee-transaction example

marcus-girolneto:
### Changes
+ Adding the NFT example with FeeTransactions support
+ Updated README
+ Added JavaDocs to the newer lifecycle functions

### Notes
+ We still need to update the tessellation version to match the version released with FeeTransactions


examples/nft-fee-transaction/gitignore
scasplte2, line 1: duplicate file of the `.gitignore`?

examples/nft-fee-transaction/project/Dependencies.scala
```diff
@@ -0,0 +1,44 @@
+import sbt.*
+
+object Dependencies {
+
+  object V {
+    val tessellation = "99.99.99" //TODO: Update with the official version with FeeTransactions
```
scasplte2, line 6: do you want to update this to a tagged version?

scasplte2 (approved):

examples/nft-fee-transaction/project/Dependencies.scala
marcus-girolneto: We don't have a fixed version in testnet, this should be updated once we migrate to intnet

marcinwadon: I think it would be `3.0.0`. but if we can, let's wait for release

PR #36, opened 2025-02-04
Initial weather metagraph structure

ramsvidor:
### Overview

This PR introduces a new metagraph example that fetches weather data from a free API and offers developers access to detailed weather reports based on location.

### Key Features

- Fetcher Daemon: Implement a daemon that fetches weather data from the weather API and pushes it on chain. We should try to fetch a large amount of data if possible but the frequency/amount may be limited by API rate limits. 

- REST API: Implement a REST API that serves historical weather data over a couple endpoints. This implementation does not need to be extensive - just enough to prove the functionality. E.g. endpoints for current weather by location + historical weather by location is enough.

- API Authentication: The API should be private and authenticate with an API token. Users can buy an API token using a DataUpdate with a FeeTransaction. API keys should also be able to be reloaded with additional API credits using a separate endpoint.


examples/weather/modules/weather-shared-data/src/main/scala/my/weathermetagraph/shared/data/lifecycle/CalculatedStateService.scala
```diff
@@ -0,0 +1,50 @@
...
+  def update(
+    snapshotOrdinal: SnapshotOrdinal,
+    state:           CalculatedState
+  ): F[Boolean] = {
+    val records = state.states.collect {
+      case WeatherCalculatedState(updates) => updates.map(_.toRecord(snapshotOrdinal))
+    }
+
+    stateRef.update { currentState =>
+      CalculatedState(currentState.states ++ state.states)
```
scasplte2, line 32: the new state given by the caller of `CalculatedStateService.update` already aggregates the previous updates of the calculated state, as this `update` function is called after the `Combiner`. So there is no need to "fold in" the updates here. I would also recommend you to move the `insertAll` operation to the `Combiner` or to the `onSnapshotConsensusResult` method. I believe this `update` function may be called during the network download process.

examples/weather/project/Dependencies.scala
```diff
@@ -0,0 +1,49 @@
+import sbt.*
+
+object Dependencies {
+
+  object V {
+    val tessellation = "2.12.0"
+    val metakit = "0.0.2"
+    val doobie = "1.0.0-RC6"
```
scasplte2, line 8: this is just a personal preference but perhaps we should standardize around a postgres lib? I believe Gabriel is using Skunk but perhaps we used doobie in the other metagraph example?

examples/weather/modules/weather-shared-data/src/main/scala/my/weathermetagraph/shared/data/domain/MetagraphState.scala
```diff
@@ -0,0 +1,140 @@
...
+    def updates: List[WeatherDataUpdate]
+  }
+
+  @derive(encoder, decoder)
+  sealed trait WeatherDataOnChainState {
+    def updates: List[WeatherDataUpdate]
+  }
+
+  @derive(encoder, decoder)
+  case class WeatherUpdate(
```
scasplte2, line 34: with these being updates, does this mean that any user can submit weather updates? How do you restrict who is allowed to push new values?

examples/weather/modules/weather-data-l1/src/main/scala/my/metagraph/data/l1/WeatherDataL1App.scala
```diff
@@ -0,0 +1,17 @@
+package my.metagraph.data.l1
+
+import my.weathermetagraph.buildinfo.BuildInfo
+import org.tessellation.currency.l1.CurrencyL1App
+import org.tessellation.schema.cluster.ClusterId
+import org.tessellation.schema.semver.{MetagraphVersion, TessellationVersion}
+
+import java.util.UUID
+
+object WeatherDataL1App
```
scasplte2, line 10: since the ML0 nodes have to do all the blockchain operations and are submitting the updates to the chain anyway, why not make the Data-L1 nodes responsible for fetching data from the external APIs? This way they are doing a little more than just presenting an endpoint for accepting updates.

examples/weather/.scalafix.conf
```diff
@@ -0,0 +1,10 @@
+rules = [
+  OrganizeImports
+]
+
+OrganizeImports {
+  groupedImports                    = Merge
+  coalesceToWildcardImportThreshold = 6
+  groups                            = ["re:javax?\\.", "cats.", "scala.", "org.tessellation.", "com.my.nft.", "*"]
```
scasplte2, line 8: might want to add the namespace you used here so that the imports format into groups

scasplte2 (dismissed):
Overall seems like a reasonable approach but I didn't see anything about the "read" portion of how users interact with the data. Will this be forthcoming?

scasplte2:
Apologies, meant to hit "Comment" not approve.

PR #38, opened 2025-10-07
 github-token is no longer required

ffox77:
removed the following line from all the example's build.sbt

 resolvers += Resolver.githubPackages("abankowski", "http-request-signer"),


nickmeessen:
we also need to remove `addSbtPlugin("com.codecommit" % "sbt-github-packages" % "0.5.3")`, everywhere. e.g. changes from https://github.com/Constellation-Labs/currency.g8/pull/15/files

AlexBrandes (changes requested):
Could you confirm that you've tested that these projects will build correctly with this change? 

My understanding was that this package can only be removed once Tessellation is upgraded to V3.x.x and we still have all these examples on 2.x.x.

nickmeessen:
can close this now, all examples have been updated to V3 which included these changes for each example 👍

